PR #7, opened 2023-06-29
AUT-101 Populate README

dariacm:


dariacm:
@kibertoad Not my best work I am afraid, I lacked some knowledge in messaging queues and the somewhat new language didn't help. I expect to learn from your review 🙏🏻

README.md
```diff
@@ -1,2 +1,65 @@
-# message-queue-toolkit
-Useful utilities, interfaces and base classes for message queue handling
+# message-queue-toolkit ✉️
+Useful utilities, interfaces and base classes for message queue handling.
+
+## Overview
+
+This is an abstraction to switch between different queue systems without having to implement your own deserialization, error handling, etc. The library provides utilities, interfaces and base classes to build the support for any queue system you may need in your service and already implements support for the following:
+
+* AMQP (Advanced Message Queuing Protocol)
```
kibertoad, line 8:
```suggestion
* AMQP 0-9-1 (Advanced Message Queuing Protocol), used e. g. by RabbitMQ
```

README.md
```diff
@@ -1,2 +1,65 @@
...
+* Configures the queue
+* Overrides the `processMessage()` method
+
+> **_NOTE:_**  See [SqsPermissionConsumer.ts](./packages/sqs/test/consumers/SqsPermissionConsumer.ts) for a practical example.
+
+## Error Handling
+
+`processMessage()` return response is `Either<'retryLater', 'success'>`, meaning that it could be an either one value or the other. In case of `retryLater`, the abstract consumer is instructed to requeue the message. Otherwise, in case of success, the message is finally removed from the queue. In case of any error, the abstract consumer will also requeue the message. When overriding the `processMessage()` method, you should leverage the possible types to process the message as you need.
+
+## Schema Validation and Deserialization
```
kibertoad, line 41: Since this is only relevant for consumers, I would adjust indentation level (`##`) so that it's a subsection.

README.md
```diff
@@ -1,2 +1,65 @@
...
+* Defines a queue name
+* Defines a message schema
+* Configures the queue
+* Overrides the `processMessage()` method
+
+> **_NOTE:_**  See [SqsPermissionConsumer.ts](./packages/sqs/test/consumers/SqsPermissionConsumer.ts) for a practical example.
+
+## Error Handling
+
+`processMessage()` return response is `Either<'retryLater', 'success'>`, meaning that it could be an either one value or the other. In case of `retryLater`, the abstract consumer is instructed to requeue the message. Otherwise, in case of success, the message is finally removed from the queue. In case of any error, the abstract consumer will also requeue the message. When overriding the `processMessage()` method, you should leverage the possible types to process the message as you need.
```
kibertoad, line 39: this is specific to consumers, I would adjust hiearchy accordingly.

README.md
```diff
@@ -1,2 +1,65 @@
...
+
+## Fan-out to Multiple Consumers
+
+SQS queues are built in a way that every message is only consumed once, and then deleted. If you want to do fan-out to multiple consumers, you need SNS topic in the middle, which is then propagated to all the SQS queues that have subscribed.
+
+> **_NOTE:_**  See [SnsPermissionPublisher.ts](./packages/sns/test/publishers/SnsPermissionPublisher.ts) and [SnsSqsPermissionConsumer.ts](./packages/sns/test/consumers/SnsSqsPermissionConsumer.ts) for a practical example.
+
+## Automatic Queue and Topic Creation
+
+The `init()` method in the abstract queue publishers and consumers calls the `assertQueue()` method. This allows to assert a queue into existence and if the queue does not exist, it will be created.
```
kibertoad, line 63: This focuses too much on implementation details, end users of the library don't care much about which methods are called.

Important part is that both publishers and consumers accept queue configuration and name as parameters, and if referenced queue does not exist at the moment of consumer or publisher instantiation, they are created.

README.md
```diff
@@ -1,2 +1,65 @@
...
+
+> **_NOTE:_**  See [SqsPermissionConsumer.ts](./packages/sqs/test/consumers/SqsPermissionConsumer.ts) for a practical example.
+
+## Error Handling
+
+`processMessage()` return response is `Either<'retryLater', 'success'>`, meaning that it could be an either one value or the other. In case of `retryLater`, the abstract consumer is instructed to requeue the message. Otherwise, in case of success, the message is finally removed from the queue. In case of any error, the abstract consumer will also requeue the message. When overriding the `processMessage()` method, you should leverage the possible types to process the message as you need.
+
+## Schema Validation and Deserialization
+
+Message deserialization is done within the abstract consumer _before_ processing the message. Deserialization is done for a message, given a message schema type. We use `zod` to define the message schema type.
```
kibertoad, line 43: I would avoid colloquial `we` in this context

README.md
```diff
@@ -1,2 +1,65 @@
...
+
+If
+* The message before deserialization is `null`
+* Deserialization returns a `MessageInvalidFormatError` error for a syntax error
+* Deserialization returns a `MessageValidationError` error for a `zod` schema validation error
+* Deserialization returns an empty value
+
+Then `abort` value is returned. In this case, the message is automatically nacked without requeueing by the abstract consumer and processing fails.
```
kibertoad, line 51: > Then `abort` value is returned. 

This is an implementation details. End users of the library only care about the second part - that message is automatically nacked in case message cannot be deserialized or does not pass the schema validation.

README.md
```diff
@@ -1,2 +1,65 @@
...
+
+`processMessage()` return response is `Either<'retryLater', 'success'>`, meaning that it could be an either one value or the other. In case of `retryLater`, the abstract consumer is instructed to requeue the message. Otherwise, in case of success, the message is finally removed from the queue. In case of any error, the abstract consumer will also requeue the message. When overriding the `processMessage()` method, you should leverage the possible types to process the message as you need.
+
+## Schema Validation and Deserialization
+
+Message deserialization is done within the abstract consumer _before_ processing the message. Deserialization is done for a message, given a message schema type. We use `zod` to define the message schema type.
+
+If
+* The message before deserialization is `null`
+* Deserialization returns a `MessageInvalidFormatError` error for a syntax error
```
kibertoad, line 47: In this context specific errors are an implementation detail, they are never propagated to the end user.

They can be mentioned if you choose to explain that you can define and pass custom `ErrorResolver` - if either of these errors are returned during error resolution, AbstractConsumer will nack the message.

README.md
kibertoad, line 8: It would be helpful to mention which dependencies need to be installed for which protocol. See package.json for each subpackage for their names (e. g. `@message-queue-toolkit/sqs`)

README.md
```diff
@@ -1,2 +1,65 @@
...
+
+* AMQP (Advanced Message Queuing Protocol)
+* SQS (Simple Queue Service)
+* SNS (Simple Notification Service)
+
+## Basic Usage
+
+### Publishers
+
+Build a publisher to send messages to a queue or a topic in a messaging system by creating a class that:
```
kibertoad, line 16: I would start this section with something along the lines of

`message-queue-toolkit provides base classes for implementing publishers for each of the supported protocol. They implement the following public methods:

constructor // explain parameters

init // explain that it needs to be invoked before publisher can be used

close // explain it needs to be invoked when stopping application

publish // explain that first parameter is the message payload, and second is protocol-dependent set of message parameters, you can also link to SQS/SNS/amqp-node documentation on their message parameters.
`

README.md
```diff
@@ -1,2 +1,65 @@
...
+
+* Extends the abstract publisher class provided for the messaging system of choice
+* Defines a queue name
```
kibertoad, line 19: publisher by itself does not define a queue name or message schema, it just accepts them as parameters; that would be most useful if described within the context of specific public methods that abstract publishers expose

README.md
```diff
@@ -1,2 +1,65 @@
...
+* Defines a message schema
+* Configures the queue
+
+> **_NOTE:_**  See [SqsPermissionPublisher.ts](./packages/sqs/test/publishers/SqsPermissionPublisher.ts) for a practical example.
+
+#### Consumers
+
+Build a recipient of messages from a queue or a topic in a messaging system by creating a class that:
```
kibertoad, line 27: See suggestions for the publisher structure

README.md
kibertoad, line 39: I would rephrase it somewhat, starting like
`When implementing message handler in consumer (by overriding `processMessage()` method), you are expected to return an instance of `Either`, containing either an error `retryLater`, or result `success`.

kibertoad, line 39: >  In case of any error

To be more precise, `if an error is thrown while processing the message`

kibertoad (approved):

kibertoad:
@dariacm Thanks, that's a great starting point! I'll polish it further later on.

PR #31, opened 2023-08-10
Barrier Pattern

CarlosGamero:
WIP

### DIFF
--- a/packages/core/index.ts
+++ b/packages/core/index.ts
@@ -21,6 +21,7 @@ export type {
   CommonQueueLocator,
   DeletionConfig,
   MultiSchemaPublisherOptions,
+  BarrierCallback,
 } from './lib/queues/AbstractQueueService'
 
 export {
--- a/packages/core/lib/queues/AbstractQueueService.ts
+++ b/packages/core/lib/queues/AbstractQueueService.ts
@@ -39,6 +39,10 @@ export type ExistingQueueOptionsMultiSchema<
 > = ExistingQueueOptions<QueueLocatorType> &
   MultiSchemaConsumerOptions<MessagePayloadSchemas, ExecutionContext>
 
+export type BarrierCallback<MessagePayloadSchema extends object> = (
+  message: MessagePayloadSchema,
+) => Promise<boolean>
+
 export type DeletionConfig = {
   deleteIfExists?: boolean
   forceDeleteInProduction?: boolean
@@ -106,6 +110,7 @@ export abstract class AbstractQueueService<
   protected abstract resolveSchema(
     message: MessagePayloadSchemas,
   ): Either<Error, ZodSchema<MessagePayloadSchemas>>
+
   protected abstract resolveMessage(
     message: MessageEnvelopeType,
   ): Either<MessageInvalidFormatError | MessageValidationError, unknown>
--- a/packages/core/lib/queues/HandlerContainer.ts
+++ b/packages/core/lib/queues/HandlerContainer.ts
@@ -1,20 +1,28 @@
 import type { Either } from '@lokalise/node-core'
 import type { ZodSchema } from 'zod'
 
-export class MessageHandlerConfig<const MessagePayloadSchemas, const ExecutionContext> {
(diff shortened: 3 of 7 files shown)


packages/amqp/lib/AbstractAmqpBaseConsumer.ts
```diff
@@ -42,8 +42,27 @@ export abstract class AbstractAmqpBaseConsumer<MessagePayloadType extends object
...
 
+  /**
+   * Override to implement barrier pattern
+   */
+  public shouldProcessMessageLater(
+    _message: MessagePayloadType,
+    _messageType: string,
+  ): Promise<boolean> {
+    return Promise.resolve(false)
+  }
```
CarlosGamero, line 53: By the default we will always process the message right away

packages/amqp/lib/AbstractAmqpBaseConsumer.ts
```diff
@@ -120,7 +136,7 @@ export abstract class AbstractAmqpBaseConsumer<MessagePayloadType extends object
         const resolvedLogMessage = this.resolveMessageLog(deserializedMessage.result, messageType)
         this.logMessage(resolvedLogMessage)
       }
-      this.processMessage(deserializedMessage.result, messageType)
+      this.internalProcessMessage(deserializedMessage.result, messageType)
```
kibertoad, line 139: why introduce additional method and not just handle the new branch in-place here? there are plenty of layers of abstraction here already 😂

packages/amqp/lib/AbstractAmqpBaseConsumer.ts
```diff
@@ -42,8 +42,24 @@ export abstract class AbstractAmqpBaseConsumer<MessagePayloadType extends object
     }
   }
 
+  /**
+   * Override to implement barrier pattern
+   */
+  public preHandlerBarrier(_message: MessagePayloadType, _messageType: string): Promise<boolean> {
+    return Promise.resolve(true)
```
kibertoad, line 49: what if it ends with an error? probably we should return an Either here as well

README.md
```diff
@@ -52,6 +52,7 @@ It consists of the following submodules:
 * `close()`, which needs to be invoked when stopping the application;
 * `processMessage()`, which accepts as parameter a `message` following a `zod` schema and should be overridden with logic on what to do with the message;
 * `start()`, which invokes `init()` and `processMessage()` and handles errors.
+* `preHandlerBarrier`, which accepts as a parameter a `message` following a `zod` schema and can be overridden to enable the build-in barrier pattern to facilitate out-of-order message handling
```
kibertoad, line 55: It's unclear how to use it. Can you add a separate doc section with an example?

CarlosGamero: I will do 👍

packages/amqp/lib/AbstractAmqpBaseConsumer.ts
CarlosGamero: I am not sure if I am getting your message, do you mean it should return a `Promise<Either<'retryLater', 'success'>>`?

if it is the case, I have no strong opinion but I wanted to make it simpler with just a boolean (`true` means that the message can be processed and false means `retryLater`). If there is an exception it is going to be handled as we are actually handling them in the `internalProcessMessage` call -> 
```
const result: Either<'retryLater' | Error, 'success'> = await this.internalProcessMessage(
          deserializedMessage.result,
          messageType,
        )
          .catch((err) => {
            // ToDo we need sanity check to stop trying at some point, perhaps some kind of Redis counter
            // If we fail due to unknown reason, let's retry
            this.handleError(err)
            return {
              // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
              error: err,
            }
          })
          .finally(() => {
            this.transactionObservabilityManager?.stop(transactionSpanId)
          })
```

kibertoad: gotcha, that should be good enough

kibertoad (approved):

PR #45, opened 2023-10-15
Explicitly set multi-schema consumer execution context

kibertoad:


kibertoad:
@drdaemos I remember you previously mentioned that implicit execution context is confusing, please take a look and let me know if it's a better approach.

CarlosGamero (approved):

PR #52, opened 2023-11-20
Prevent flakiness

kibertoad:
This started as a dependency updated, but then I noticed that quite a few tests are flaky. I've investigated the reasons, and this resulted in several improvements across the board. 

* Automatic queue recreation is no longer flaky, as it properly awaits deletion of existing queue;
* Logging improved and includes SQS consumer queue;
* One big discovery: SQS Client shouldn't be disposed of at all, doing so prevents consumers from closing down properly in some cases, and not closing it at all is perfectly fine.
* Many cases of message-queue-toolkit test flakiness fixed via proper consumer closing;


packages/sqs/test/consumers/SqsPermissionsConsumerMonoSchema.spec.ts
```diff
@@ -98,16 +101,15 @@ describe('SqsPermissionsConsumerMonoSchema', () => {
         QueueUrl: diContainer.cradle.permissionPublisher.queueUrl,
       })
       const reply = await sqsClient.send(command)
-      expect(reply.Messages).toBeUndefined()
+      expect(reply.Messages!.length).toBe(0)
```
kibertoad, line 104: reply signature changed in newer SDK version

packages/sqs/test/consumers/SqsPermissionsConsumerMonoSchema.spec.ts
```diff
@@ -193,38 +195,5 @@ describe('SqsPermissionsConsumerMonoSchema', () => {
         expect(usersPermissions[0]).toHaveLength(2)
       })
     })
-
-    describe('error handling', () => {
```
kibertoad, line 197: moved to a separate file

packages/sns/test/utils/testContext.ts
```diff
@@ -58,8 +58,6 @@ export async function registerDependencies(
       },
       {
         lifetime: Lifetime.SINGLETON,
-        asyncDispose: 'destroy',
-        asyncDisposePriority: 40,
```
CarlosGamero, line 62: Mmm, really interesting and good to know! It might be causing flakiness problems in library clients, right?

packages/sqs/lib/utils/sqsUtils.ts
```diff
@@ -106,7 +112,24 @@ export async function deleteQueue(client: SQSClient, queueName: string) {
...
+            QueueNamePrefix: queueName,
+          }),
+        )
+        return !queueList.QueueUrls || queueList.QueueUrls.length === 0
+      })
+    }
   } catch (err) {
+    // This is fine
+    // @ts-ignore
+    if (err.name === 'QueueDoesNotExist') {
```
CarlosGamero, line 129: 🟢 Should this be a constant?

packages/sqs/test/fakes/FakeConsumerErrorResolver.ts
```diff
@@ -1,15 +1,18 @@
 import { SqsConsumerErrorResolver } from '../../lib/errors/SqsConsumerErrorResolver'
 
 export class FakeConsumerErrorResolver extends SqsConsumerErrorResolver {
+  public errors: Error[]
   public handleErrorCallsCount: number
```
CarlosGamero, line 5: 🟢 I think that with `errors`,  `handleErrorCallsCount` is redundant

packages/sqs/test/fakes/FakeConsumerErrorResolver.ts
```diff
@@ -1,15 +1,18 @@
...
   constructor() {
     super()
 
     this.handleErrorCallsCount = 0
+    this.errors = []
```
CarlosGamero, line 10: 🟢 Would make sense to add a `clean` method?

packages/sqs/test/fakes/FakeConsumerErrorResolver.ts
```diff
@@ -1,15 +1,18 @@
...
   }
 
   public override processError(error: unknown) {
     this.handleErrorCallsCount++
+    this.errors.push(error as Error)
```
CarlosGamero, line 15: 🟢 Wondering if we should control the case of error not being of type Error

CarlosGamero (approved):
Outstanding job Igor 🚀! thank you so much for doing it. I left some small comments but they are not blockers so feel free to ignore them if you think they don't make sense

packages/sns/test/utils/testContext.ts
kibertoad: yeah, I would expect so. we should also do this change on autopilot and check if it also works fine there

packages/sqs/test/fakes/FakeConsumerErrorResolver.ts
kibertoad: I wouldn't do it until there is a real case of that being a problem, and that has never happened in my experience in any library so far

packages/sqs/test/fakes/FakeConsumerErrorResolver.ts
kibertoad: fair point, we can just check the length

PR #60, opened 2023-12-06
Implement handler spy for easy awaiting of expected messages

kibertoad:
This is not yet used by any consumers, to be done in follow-up PRs. 
This PR is to validate the API that developers would be using.


packages/core/lib/queues/HandlerSpy.ts
```diff
@@ -0,0 +1,115 @@
...
+      (!processingResult || spyResult.processingResult === processingResult)
+    )
+  }
+
+  waitForEvent(
+    fields: Partial<MessagePayloadSchemas>,
+    processingResult?: MessageProcessingResult,
+  ): Promise<SpyResult<MessagePayloadSchemas>> {
+    // @ts-ignore
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
kibertoad, line 58: this needs to be fixed upstream in caching library, it currently types `items` as `any` :-/

packages/core/lib/queues/HandlerSpy.ts
```diff
@@ -0,0 +1,115 @@
...
+    )
+    if (processedMessageEntry) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+      return Promise.resolve(processedMessageEntry.value)
+    }
+
+    let resolve: (
+      value: SpyResult<MessagePayloadSchemas> | PromiseLike<SpyResult<MessagePayloadSchemas>>,
+    ) => void
+    const spyPromise = new Promise<SpyResult<MessagePayloadSchemas>>((_resolve) => {
```
kibertoad, line 73: this can be removed after next version of TypeScript lands, it's supposed to have https://github.com/tc39/proposal-promise-with-resolvers

PR #92, opened 2024-02-22
Automatically update SQS queue attributes

kibertoad:


packages/sqs/lib/utils/sqsInitter.ts
```diff
@@ -1,4 +1,6 @@
 import type { SQSClient } from '@aws-sdk/client-sqs'
+import { SetQueueAttributesCommand } from '@aws-sdk/client-sqs'
+import type { QueueAttributeName } from '@aws-sdk/client-sqs/dist-types/models/models_0'
```
jmontoyaa, line 3: Is this import is correct? Should this include `/dist-types/models/models_0`?

kibertoad: great catch! autoimport went wrong.
there were more of cases like this, fixed

jmontoyaa (approved):

packages/core/.eslintrc.json
```diff
@@ -21,7 +21,7 @@
     "@typescript-eslint/no-empty-interface": "warn",
     "@typescript-eslint/ban-ts-comment": "off",
     "@typescript-eslint/no-use-before-define": "off",
-    "@typescript-eslint/no-non-null-assertion": "warn",
+    "@typescript-eslint/no-non-null-assertion": "off",
```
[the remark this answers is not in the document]
kibertoad: fixed

packages/sqs/.eslintrc.json
```diff
@@ -23,7 +23,7 @@
     "@typescript-eslint/no-empty-interface": "warn",
     "@typescript-eslint/ban-ts-comment": "off",
     "@typescript-eslint/no-use-before-define": "off",
-    "@typescript-eslint/no-non-null-assertion": "warn",
+    "@typescript-eslint/no-non-null-assertion": "off",
```
CarlosGamero, line 26: Looks like we have the same here, maybe I am missing something and this is needed so please let me know if it is the case 🙇

kibertoad: will address in a follow-up PR

CarlosGamero (approved):

PR #96, opened 2024-02-23
Avoid redundant updates if changed attributes are a strict subset of existing attributes

kibertoad:


packages/sqs/lib/utils/sqsUtils.ts
```diff
@@ -107,8 +107,7 @@ export async function assertQueue(
         throw new Error('Queue ARN was not set')
       }
 
-      if (shallowEqual(existingAttributes.result, queueConfig.Attributes)) {
-      } else {
+      if (!isShallowSubset(existingAttributes.result, queueConfig.Attributes)) {
```
CarlosGamero, line 110: is `shallowEqual` still used? I am not sure, just wondering if it could be removed

CarlosGamero (approved):

packages/sqs/lib/utils/sqsUtils.ts
kibertoad: good point, removed

PR #98, opened 2024-02-23
Avoid potential racing condition

kibertoad:


packages/sns/lib/sns/AbstractSnsService.ts
```diff
@@ -85,9 +85,13 @@ export abstract class AbstractSnsService<
   // @ts-ignore
   public topicArn: string
 
+  private isInitted: boolean
```
kibertoad, line 88: to be more explicit and not rely on implementation detail

PR #99, opened 2024-03-01
Add logging for processed messages

kibertoad:


CarlosGamero (approved):

packages/sns/lib/utils/snsInitter.ts
```diff
@@ -121,7 +121,7 @@ export async function deleteSnsSqs(
     deletionConfig.waitForConfirmation !== false,
   )
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
CarlosGamero, line 123: 🟢 I guess the aslant-disable is no longer needed

CarlosGamero: Actually, TS build is failing due to this change I think

PR #117, opened 2024-04-01
Dead letter queue

CarlosGamero:
A dead letter queue is a queue where messages that cannot be processed successfully are stored. It serves as a holding area for messages that have encountered errors or exceptions during processing, allowing developers to review and handle them later. To create a dead letter queue, you need to specify the `deadLetterQueue` parameter within `options` on the consumer configuration. The parameter should contain the following fields:

- `creationConfig`: configuration for the queue to create, if one does not exist. Should not be specified together with the   `locatorConfig`
- `locatorConfig`: configuration for resolving existing queue. Should not be specified together with the `creationConfig`
- `redrivePolicy`: an object that contains the following fields:
  - `maxReceiveCount`: the number of times a message can be received before being moved to the DLQ.

**Note:** AMQP does not support DLQ at this point, this will be handled in a future version.

**Known issue**
If you use the barrier pattern together with a dead letter queue, you should be aware that the library does not simply return the message to the queue, but creates a new one instead to avoid exhausting DLQ limits early. 
Due to this fact, in those cases, you could have duplicated messages if a service crashes after receiving `retryLater` result from the barrier and creating a new message for the retry, but before the original one is consumed (those are not being done as an atomic operation). 
In a future release, we will implement deduplication to remove this risk but in the meantime, please keep this in mind.


packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -18,24 +19,51 @@ import { Consumer } from 'sqs-consumer'
...
+      | {
+          queueNameSuffix?: string
+          QueueName?: never
+        }
+      | {
+          queueNameSuffix?: never
+          QueueName?: string
+        }
+    )
 }
```
CarlosGamero, line 48: I am not sure if the inclusion of `queueNameSuffix` and the default name for DLQ is worth it, I mean, it complicates types and name handling but from the client side there is not much benefit, what do you think?

kibertoad: it seems to be common to have one dlq per normal queue, though: https://stackoverflow.com/questions/58469216/how-many-dead-letter-queues-should-i-have-on-aws

CarlosGamero: I think I am going to remove the default DLQ naming for now, sharing my reasons to open the discussions in case you disagree.
- Not clear if it is worth it or not
- Avoid adding complexity (types and name handling)
- If later we see it makes sense, adding it is not a breaking change (but removal is)

I am more than open to discuss and continue with the default naming if it makes sense :)

kibertoad: how would you handle one-dlq-per-queue pattern without it? can you PoC it within a test consumer to check how convenient it is?

CarlosGamero: My message was not clear sorry, my idea is to drop the default name behaviour and `queueNameSuffix` and replace it by requesting either `creationConfig` or `locatorConfig`, something like:
```
type DeadLetterQueueOptions<
  CreationConfigType extends SQSCreationConfig,
  QueueLocatorType extends SQSQueueLocatorType,
> = {
  redrivePolicy: { maxReceiveCount: number }
} & (
  | {
      creationConfig: CreationConfigType
      locatorConfig?: never
    }
  | {
      creationConfig?: never
      locatorConfig: QueueLocatorType
    }
)
```

So the user will always have to define the DLQ config and we support both cases:
- DLQ per normal queue -> User will specify different queue names or locators
- Single DLQ for all queues -> User will specify the same DLQ config
(Event mixed cases where a DLQ can be shared for different queues)

kibertoad: Sure, that works, and I see why you prefer that way - we can reuse existing entities directly that way.

packages/amqp/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@message-queue-toolkit/amqp",
-  "version": "12.0.1",
+  "version": "12.0.2",
```
CarlosGamero, line 3: Not sure if `patch` or `minor`, the new parameter is there but it can't be used so I guess it is a patch

packages/amqp/test/consumers/AmqpPermissionConsumer.spec.ts
```diff
@@ -15,6 +15,52 @@ import { registerDependencies, SINGLETON_CONFIG } from '../utils/testContext'
 import { AmqpPermissionConsumer } from './AmqpPermissionConsumer'
 
 describe('AmqpPermissionConsumer', () => {
+  describe('init', () => {
```
CarlosGamero, line 18: Adding some tests to make sure unsupported options are throwing an error and not silently accepted

packages/core/lib/types/queueOptionsTypes.ts
```diff
@@ -35,15 +36,13 @@ export type DeletionConfig = {
...
   creationConfig: CreationConfigType
-} & CommonQueueOptions
+}
 
 type ExistingQueueOptions<QueueLocatorType extends object> = {
   locatorConfig: QueueLocatorType
-  deletionConfig?: DeletionConfig
   creationConfig?: never
-} & CommonQueueOptions
+}
```
CarlosGamero, line 45: `CommonQueueOptions` is already being used on `QueueOptions`, so avoiding repetition

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -18,24 +20,17 @@ import { Consumer } from 'sqs-consumer'
...
 
-export type ExtraSQSCreationParams = {
-  topicArnsWithPublishPermissionsPrefix?: string
-  updateAttributesIfExists?: boolean
-}
-
-export type SQSCreationConfig = {
-  queue: CreateQueueRequest
-  updateAttributesIfExists?: boolean
-} & ExtraSQSCreationParams
```
CarlosGamero, line 38: Moving them to service as they are common for publisher and consumer

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -159,13 +192,9 @@ export abstract class AbstractSqsConsumer<
           messageType,
         )
           .catch((err) => {
-            // ToDo we need sanity check to stop trying at some point, perhaps some kind of Redis counter
-            // If we fail due to unknown reason, let's retry
```
CarlosGamero, line 163: I guess DLQ is addressing this TODO

README.md
```diff
@@ -155,6 +157,19 @@ If the barrier method returns `false`, message will be returned into the queue f
 
 > **_NOTE:_**  See [SqsPermissionConsumer.ts](./packages/sns/test/consumers/SnsSqsPermissionConsumer.ts) for a practical example.
 
+### Dead Letter Queue
+A dead letter queue is a queue where messages that cannot be processed successfully are stored. It serves as a holding area for messages that have encountered errors or exceptions during processing, allowing developers to review and handle them later.
+To create a dead letter queue, you need to specify the `deadLetterQueue` parameter within options on the consumer configuration. The parameter should contain the following fields:
+- `creationConfig`: configuration for queue to create, if one does not exist. Should not be specified together with the `locatorConfig`
+- `locatorConfig`: configuration for resolving existing queue. Should not be specified together with the `creationConfig`
+- `redrivePolicy`: an object that contains the following fields:
```
kibertoad, line 165: this is very SQS-specific, I wonder if we should have a higher-level concept on top of it

README.md
```diff
@@ -155,6 +157,19 @@ If the barrier method returns `false`, message will be returned into the queue f
...
+  - `maxReceiveCount`: the number of times a message can be received before being moved to the DLQ.
+
+> **_NOTE:_**  DLQ is not available for AMQP consumers.
+
+> **_NOTE:_**  DLQ could produce message duplication if used together with the barrier pattern (due to the 
+    retry mechanism), although this is unlikely to happen in practice.
```
kibertoad, line 171: I would clarify it - if a service crashes after receiving retryLater result from a barrier and creating a new message for the retry, but before the original one is consumed, there will be a duplicate, because those are not being done as an atomary operation.
It would also be helpful to clarify how the barrier works when used together with DLQ, that we do not simply return message to the queue, but create a new one instead, in order to avoid exhausting DLQ limits early.

packages/core/lib/types/queueOptionsTypes.ts
```diff
@@ -59,6 +58,14 @@ export type QueuePublisherOptions<
   messageSchemas: readonly ZodSchema<MessagePayloadSchemas>[]
 }
 
+export type DeadLetterQueueOptions<
+  CreationConfigType extends object,
+  QueueLocatorType extends object,
+> = {
+  redrivePolicy: { maxReceiveCount: number }
```
kibertoad, line 65: Wonder if this can be a general interface, or if it should be placed as a SQSDeadLetterQueueOptions entity. We need to look into what kind of settings we would need for AMQP in order to understand what a good general interface could be.

CarlosGamero: Mmm yeah, you are right,  I don't see any kind of "max allowed retries" option on AMQP
> There are three identified situations where a message becomes undeliverable after reaching RabbitMQ:
> - A message is negatively acknowledged by the consumer
> - The TTL of a message expires
> - The queue reaches capacity

But I guess we want to have several retries for AMQP too right? So I guess we should handle it by ourselves.

I am wondering if we can keep this same API for both and a generic interface for extra properties like:
```
{
  redrivePolicy: { maxReceiveCount: number } & ModuleExtraParams
}
```

but I am not really sure, I would appreciate your thoughts 🙏

README.md
CarlosGamero: Will make it more generic once we address the comment below, thanks Igor 🙇

README.md
CarlosGamero: Makes sense, I will apply your suggestions to the node

CarlosGamero: Done, please have a look if it is clearer now 🙏

packages/core/lib/types/queueOptionsTypes.ts
CarlosGamero: Discussed on Slack, for visibility:

`redrivePolicy` has been extracted to be SQS specific because even if we finally implement retries on AMQP (TBD), `redrivePolicy` is very couple to SQS and we want to have more flexibility to accommodate the decision we make for AMQP.

README.md
CarlosGamero: Added a note a top explaining DLQ support, so I guess it is fine to keep this coupled to SQS for now

kibertoad (approved):
Great job, Carlos, this is epic!

PR #124, opened 2024-04-22
Visibility timeout heartbeat 

CarlosGamero:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html

### DIFF
--- a/packages/sns/package.json
+++ b/packages/sns/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@message-queue-toolkit/sns",
-  "version": "12.1.1",
+  "version": "13.0.0",
   "private": false,
   "license": "MIT",
   "description": "SNS adapter for message-queue-toolkit",
@@ -25,7 +25,7 @@
     "prepublishOnly": "npm run build:release"
   },
   "dependencies": {
-    "@lokalise/node-core": "^9.10.1",
+    "@lokalise/node-core": "^9.14.0",
     "sqs-consumer": "^9.1.0",
     "zod": "^3.22.4"
   },
@@ -33,7 +33,7 @@
     "@aws-sdk/client-sns": "^3.476.0",
     "@aws-sdk/client-sqs": "^3.476.0",
     "@message-queue-toolkit/core": "^10.1.1",
-    "@message-queue-toolkit/sqs": "^12.1.1"
+    "@message-queue-toolkit/sqs": "^13.0.0"
   },
   "devDependencies": {
     "@aws-sdk/client-sns": "^3.529.1",
--- a/packages/sns/test/consumers/SnsSqsPermissionConsumer.spec.ts
+++ b/packages/sns/test/consumers/SnsSqsPermissionConsumer.spec.ts
@@ -3,7 +3,7 @@ import type { SQSClient } from '@aws-sdk/client-sqs'
 import { waitAndRetry } from '@lokalise/node-core'
 import { assertQueue, deleteQueue, getQueueAttributes } from '@message-queue-toolkit/sqs'
 import type { AwilixContainer } from 'awilix'
-import { asFunction } from 'awilix'
+import { asValue, asFunction } from 'awilix'
 import { describe, beforeEach, afterEach, expect, it, beforeAll } from 'vitest'
 
 import { assertTopic, deleteTopic } from '../../lib/utils/snsUtils'
@@ -449,8 +449,8 @@ describe('SnsSqsPermissionConsumer', () => {
 
     beforeEach(async () => {
       diContainer = await registerDependencies({
-        permissionConsumer: asFunction(() => undefined),
-        permissionPublisher: asFunction(() => undefined),
+        permissionConsumer: asValue(() => undefined),
+        permissionPublisher: asValue(() => undefined),
       })
     })
 
--- a/packages/sns/test/consumers/SnsSqsPermissionConsumer.spec.ts
+++ b/packages/sns/test/consumers/SnsSqsPermissionConsumer.spec.ts
@@ -1,11 +1,13 @@
 import type { SNSClient } from '@aws-sdk/client-sns'
 import type { SQSClient } from '@aws-sdk/client-sqs'
+import { waitAndRetry } from '@lokalise/node-core'
 import { assertQueue, deleteQueue, getQueueAttributes } from '@message-queue-toolkit/sqs'
 import type { AwilixContainer } from 'awilix'
+import { asFunction } from 'awilix'
 import { describe, beforeEach, afterEach, expect, it, beforeAll } from 'vitest'
 
 import { assertTopic, deleteTopic } from '../../lib/utils/snsUtils'
-import type { SnsPermissionPublisher } from '../publishers/SnsPermissionPublisher'
+import { SnsPermissionPublisher } from '../publishers/SnsPermissionPublisher'
 import { registerDependencies } from '../utils/testContext'
 import type { Dependencies } from '../utils/testContext'
 
@@ -88,6 +90,7 @@ describe('SnsSqsPermissionConsumer', () => {
             QueueName: 'existingQueue',
             Attributes: {
               KmsMasterKeyId: 'othervalue',
+              VisibilityTimeout: '10',
             },
           },
           updateAttributesIfExists: true,
@@ -106,7 +109,10 @@ describe('SnsSqsPermissionConsumer', () => {
         queueUrl: newConsumer.subscriptionProps.queueUrl,
       })
 
-      expect(attributes.result?.attributes!.KmsMasterKeyId).toBe('othervalue')
+      expect(attributes.result?.attributes).toMatchObject({
+        KmsMasterKeyId: 'othervalue',
+        VisibilityTimeout: '10',
+      })
     })
 
     it('updates existing queue when one with different attributes exist and sets the policy', async () => {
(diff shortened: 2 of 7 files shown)


packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -54,8 +55,9 @@ export type SQSConsumerOptions<
   SQSCreationConfig,
   SQSQueueLocatorType
 > & {
-  consumerOverrides?: Partial<ConsumerOptions>
+  consumerOverrides?: Pick<ConsumerOptions, 'terminateVisibilityTimeout' | 'heartbeatInterval'>
```
CarlosGamero, line 58: `Partial<ConsumerOptions>` was allowing to override `handler`, `queueUrl` or `sqsClient` which is not good from my point of view.

I think that it is a good idea to not just expose everything without knowing how it works, so just exposing what we are using but please feel free to contradict me if you disagree

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -170,12 +172,21 @@ export abstract class AbstractSqsConsumer<
 
   public async start() {
     await this.init()
+    if (this.consumer) this.consumer.stop()
+
+    const visibilityTimeout = await getQueueAttributes(
+      this.sqsClient,
+      { queueUrl: this.queueUrl },
+      ['VisibilityTimeout'],
+    )
```
CarlosGamero, line 181: `VisibilityTimeout` should be picked from the queue to support `queueLocator` case

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
kibertoad: yeah, fair point.
are there any other potentially useful options?

CarlosGamero: Not sure if there are other useful options, do you think that makes sense to spend time researching other options now? 

The first version I wrote was using `Omit` -> https://github.com/kibertoad/message-queue-toolkit/pull/124/commits/e1bca7679e8ddcc0ddcadf6eba9011b30a19b853

but after giving it a thought I changed it because I don't know how other parameters behave and that can overlap with other things (eg: `visibilityTimeout`). My reasoning for choosing `Pick` is that I don't want to spend time now investigating all possibilities, and adding them later if needed is a minor release, but removing an overlap option is a major one.

But if you still think that Omit is a better option I will change it, I mean I see the benefits of exposing all possibilities to not have to work on a new version if the parameter is already there

kibertoad: I would prefer to provide users an escape hatch for cases when SQS/SNS support something that we currently don't, and not force them to rely on us to expose something for them. So I would only hide the options that we rely upon, that might break if someone overrides them on top of our own dynamic setup.

I'm not even sure about hiding `visibilityTimeout`. Isn't it a perfectly legit use-case for someone to choose not to use the refresh with the heartbeat, but just to have some explicitly set static timeout?

packages/sns/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@message-queue-toolkit/sns",
-  "version": "12.1.1",
+  "version": "13.0.0",
```
kibertoad, line 3: before we release this version, let's also merge https://github.com/kibertoad/message-queue-toolkit/pull/121, so that we don't need to do semver majors in a row

packages/sns/test/consumers/SnsSqsPermissionConsumer.spec.ts
```diff
@@ -435,4 +441,71 @@ describe('SnsSqsPermissionConsumer', () => {
...
+      const consumer1 = new SnsSqsPermissionConsumer(diContainer.cradle, {
+        creationConfig: {
+          topic: { Name: topicName },
+          queue: { QueueName: queueName, Attributes: { VisibilityTimeout: '2' } },
+        },
+        consumerOverrides: { heartbeatInterval: heartbeatEnabled ? 1 : undefined },
+        removeHandlerOverride: async () => {
+          consumer1IsProcessing = true
+          // wait for consumer2 to process message while this is still processing
+          await waitAndRetry(() => consumer2Counter > 0, 100, 30)
```
kibertoad, line 476: we probably should also store the result of this `waitAndRetry` and assert it resolved to undefined, if intention is to ensure it never processes. or the idea is that it just takes long? then we should assert it returns true :D

packages/sns/test/consumers/SnsSqsPermissionConsumer.spec.ts
```diff
@@ -435,4 +441,71 @@ describe('SnsSqsPermissionConsumer', () => {
...
+        locatorConfig: { topicArn: consumer1.subscriptionProps.topicArn },
+      })
+
+      await publisher.publish({ id: '10', messageType: 'remove' })
+      // wait for consumer1 to start processing to start second consumer
+      await waitAndRetry(() => consumer1IsProcessing, 5, 5)
+      await consumer2.start()
+
+      // wait for both consumers to process message
+      await waitAndRetry(() => consumer1Counter > 0 && consumer2Counter > 0, 100, 40)
```
kibertoad, line 505: we should assert that it resolves to true and not just timeouts

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
kibertoad: Can we avoid this extra lookup in case we are not using `queueLocator`, so that we do not do extra hop? If we are creating a new queue anyway, we can rely on its visibility timeout.

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -170,12 +172,21 @@ export abstract class AbstractSqsConsumer<
...
 
-    if (this.consumer) {
-      this.consumer.stop()
-    }
     this.consumer = Consumer.create({
+      sqs: this.sqsClient,
       queueUrl: this.queueUrl,
+      visibilityTimeout: visibilityTimeout.result?.attributes?.VisibilityTimeout
```
kibertoad, line 186: 🟢 it might read a bit better if we resolve visibilityTimeout inside a variable first, and then only pass here

packages/sns/package.json
CarlosGamero: Sounds great to me 🙇 I will wait to merge this PR

packages/sns/test/consumers/SnsSqsPermissionConsumer.spec.ts
```diff
@@ -435,4 +441,71 @@ describe('SnsSqsPermissionConsumer', () => {
...
   })
+
+  describe('visibility timeout', () => {
+    const topicName = 'myTestTopic'
+    const queueName = 'myTestQueue'
+    let diContainer: AwilixContainer<Dependencies>
+
+    beforeEach(async () => {
+      diContainer = await registerDependencies({
+        permissionConsumer: asFunction(() => undefined),
```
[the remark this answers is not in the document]
CarlosGamero: Good point! will change it, thanks 🙇

packages/sns/package.json
kibertoad: feel free to merge, just don't hurry releasing it :D.
After you are done, I will address any conflicts if there are on the other PR, and release both changes

packages/sns/test/consumers/SnsSqsPermissionConsumer.spec.ts
CarlosGamero: At this point, I am only forcing this wait to give time to consumer2 to process the message, depending on the test it will resolves to `true` or `false`, I am checking it at the end of the test by using the counter.

The reason why I don't check it here is because I prefer to have the Given-When-Then structure and also to reduce the number of `heatbeatIntervalEnabled` checks in the test

packages/sns/test/consumers/SnsSqsPermissionConsumer.spec.ts
CarlosGamero: The check is done using counters, in the second `each` case I don't only want to know it is `false`, we should check that consumer1 was de only who received the message

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
CarlosGamero: yeah 😓 sorry, will apply the change 🙇

packages/sns/test/consumers/SnsSqsPermissionConsumer.spec.ts
kibertoad: if we don't care about the actual condition, maybe we should just do `await setTimeout(fixedTime)` (from `timers/promises`) then?

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
CarlosGamero: `visibilityTimeout` on consumer doesn't update the queue `VisibilityTimeout`, so I decided to hide it because both values should be in sync and is better to just pick the queue value to have a less confusing API, but please let me know if there is something I missed.

> I would prefer to provide users an escape hatch for cases when SQS/SNS support something that we currently don't, and not force them to rely on us to expose something for them.

I will change it to use Omit then, thanks for sharing Igor!

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
CarlosGamero: yeah, will add that piece of logic 🙇

packages/sns/package.json
CarlosGamero: Ah, right, this repo doesn't have auto-release 😅 sorry, I was confused, thanks for the clarification 🙏

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
kibertoad: Fair point, let's hide it from consumerOverrides and expect users to use our own option for this

kibertoad (approved):
An important improvement!

packages/sns/test/consumers/SnsSqsPermissionConsumer.spec.ts
CarlosGamero: Fair point! I will modify it

PR #128, opened 2024-04-24
Bugfix -> SNS consumer queue name is always undefined

CarlosGamero:


packages/sns/lib/utils/snsInitter.ts
```diff
@@ -58,12 +61,12 @@ export async function initSnsSqs(
...
       subscriptionArn,
       topicArn,
+      queueName: creationConfig.queue.QueueName,
       queueUrl,
     }
   }
 
-  // Check for existing resources, using the locators
-
+  // Check for existing rsources, using the locators
```
kibertoad, line 69:
```suggestion
  // Check for existing resources, using the locators
```

kibertoad (approved):

packages/sns/lib/utils/snsInitter.ts
CarlosGamero: 🤦‍♂️ thanks Igor 🙌

PR #133, opened 2024-05-01
Add type-safe domain event emitter and event registry

kibertoad:


kibertoad:
@dariacm @CarlosGamero I've merged this by mistake, please review it when you have some time!

packages/core/lib/messages/baseMessageSchemas.ts
```diff
@@ -0,0 +1,22 @@
+import z from 'zod'
+
+export const BASE_MESSAGE_SCHEMA = z.object({
```
kibertoad, line 3: @dariacm @CarlosGamero @Dosexe 
I've tried to revise the original structure that we used for defining our events previously to make it more straightforward, please review this change carefully. Migrating from one structure to another one is a major pain, let's ensure we've got this really right.

PR #134, opened 2024-05-01
Domain event emitter V2

kibertoad:


packages/core/lib/messages/baseMessageSchemas.ts
```diff
@@ -0,0 +1,25 @@
+import z from 'zod'
+
+export const BASE_MESSAGE_SCHEMA = z.object({
+  id: z.string().describe('event unique identifier'),
+  type: z.literal<string>('<replace.me>').describe('event type name'),
+  payload: z.optional(z.object({})).describe('event payload based on type'),
+})
+
+export const EXTENDED_MESSAGE_SCHEMA = BASE_MESSAGE_SCHEMA.extend({
```
kibertoad, line 9: this is the most important part to review and criticize

packages/core/lib/messages/baseMessageSchemas.ts
```diff
@@ -0,0 +1,25 @@
...
+  metadata: z
+    .object({
+      schemaVersion: z.string().min(1).describe('base event schema version'),
+      timestamp: z.string().datetime().describe('iso 8601 datetime'),
+      producerApp: z.string().min(1).describe('app/service that produced the event'),
```
Dosexe, line 14: does this field raise the visibility somehow? we will see this information only in the logs potentially

packages/core/lib/messages/baseMessageSchemas.ts
```diff
@@ -0,0 +1,25 @@
...
+      correlationId: z
+        .string()
+        .describe('unique identifier passed to all events in workflow chain'),
+    })
+    .describe('event metadata'),
+})
+
+export type BaseMessageType = z.infer<typeof BASE_MESSAGE_SCHEMA>
+
+export type MessageMetadata = 'id' | 'type' | 'metadata'
```
Dosexe, line 25: where this is used?

packages/core/lib/messages/baseMessageSchemas.ts
```diff
@@ -0,0 +1,25 @@
+import z from 'zod'
+
+export const BASE_MESSAGE_SCHEMA = z.object({
```
Dosexe, line 3: do we have plans to use `BASE_MESSAGE_SCHEMA` without metadata? If no, I would make it private

kibertoad: This is an opensource lib, so it would be great to leave space for people to use it with whatever data structure that they have or want to have. BASE_MESSAGE_SCHEMA offers the minimum subset that message-queue-toolkit itself relies on.

packages/core/lib/messages/baseMessageSchemas.ts
kibertoad: it wasn't, so I removed it. will reintroduce if something will need it in the future

packages/core/lib/messages/baseMessageSchemas.ts
kibertoad: addressed

packages/core/lib/messages/baseMessageSchemas.ts
kibertoad: > Shouldn't the specific schema extension be part of apps using the library?

Yes, definitely. Which is why there is an _optional_ extended schema for those who want a reasonable default (e. g. us), but the library itself doesn't rely on it and supports having your own.

If we ever decide to change it, I would expect we would be adding it as a separate schema, e.  g. `EXTENDED_MESSAGE_SCHEMA_V2`, so that those who are happy with the original one, can keep using it.

CarlosGamero: Yeah, the point was that I am not sure if the library should just define the minimum it needs to work and be open to extensions (it is already) without providing things the users can define based on their use cases, and are potential points of change. Also, to avoid having `EXTENDED_MESSAGE_SCHEMA_V{{N}}` schemas for backward compatibility.

I think in our case, I would go with creating our own extension (could be `MY_SCHEMA = EXTENDED_MESSAGE_SCHEMA`) to have flexibility and not have to modify the library to just modify the schema.

But as I mentioned, I am not sure if this concern is valid or if it is stupid, so if you think it is fine, I am also fine and  please feel free to resolve this thread if it is the case

kibertoad: > I think in our case, I would suggest going with creating our own extension (could be MY_SCHEMA = EXTENDED_MESSAGE_SCHEMA) to have flexibility and not have to modify the library to just modify the schema (same problem we have with WS rooms, where to add a room we need a new library version)

Very fair point. But also for standartization sake I would like to have one single place where our common schema is defined, so that each service doesn't have to do it over and over again. It doesn't have to be this library, though, we could put it e. g. in `shared-ts-libs` as a part of `events-common`.

I would still provide EXTENDED_MESSAGE_SCHEMA as an opinionated starting point for people/teams who might benefit from it, but won't couple to it anywhere in message-queue-toolkit itself.

I agree that when we need to adjust our schema, we should be able to do it outside of message-queue-toolkit.

CarlosGamero (commented):
LGTM 🚀! thanks, Igor for doing it and also answering my questions and concerns 🧡

CarlosGamero (approved):

Dosexe (approved):

PR #141, opened 2024-05-09
Messages barrier max time

CarlosGamero:
**Description**
If a message fails to be processed, and is there another message waiting for it on the barrier pattern, it will be there waiting forever.

**Solution:** 
- Add max time a message can be waiting with the `retryLater` feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ They implement the following public methods:
         * `messageTypeField` - which field in the message describes the type of a message. This field needs to be defined as `z.literal` in the schema and is used for resolving the correct schema for validation
         * `locatorConfig` - configuration for resolving existing queue and/or topic. Should not be specified together with the `creationConfig`.
         * `creationConfig` - configuration for queue and/or topic to create, if one does not exist. Should not be specified together with the `locatorConfig`.
-        * `messageTimestampField` - which field in the message contain the message creation date (by default it is `timestamp`). This field needs to be a `Date` object or ISO-8601 date string, if your messages doesn't contain it the library will add one automatically to avoid infinite loops on consumer;
+        * `messageTimestampField` - which field in the message contains the message creation date (by default it is `timestamp`). This field needs to be a `Date` object or ISO-8601 date string, if your message doesn't contain it the library will add one automatically to avoid infinite loops on consumer;
 * `init()`, prepare publisher for use (e. g. establish all necessary connections);
 * `close()`, stop publisher use (e. g. disconnect);
 * `publish()`, send a message to a queue or topic. It accepts the following parameters:
--- a/README.md
+++ b/README.md
@@ -171,9 +171,7 @@ To create a dead letter queue, you need to specify the `deadLetterQueue` paramet
 - `redrivePolicy`: an object that contains the following fields:
   - `maxReceiveCount`: the number of times a message can be received before being moved to the DLQ.
 
-If you use the barrier pattern together with a dead letter queue, you should be aware that the library does not simply return the message to the queue, but creates a new one instead to avoid exhausting DLQ limits early.
-Due to this fact, in those cases, you could have duplicated messages if a service crashes after receiving `retryLater` result from the barrier and creating a new message for the retry, but before the original one is consumed (those are not being done as an atomic operation).
-In a future release, we will implement deduplication to remove this risk but in the meantime, please keep this in mind.
+> **_NOTE:_**  if a message is stuck returning retryLater, it will be moved to the DLQ after the `maxRetryDuration` is reached.
 
 ## Fan-out to Multiple Consumers
 
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ They implement the following public methods:
         * `messageTypeField` - which field in the message describes the type of a message. This field needs to be defined as `z.literal` in the schema and is used for resolving the correct schema for validation
         * `locatorConfig` - configuration for resolving existing queue and/or topic. Should not be specified together with the `creationConfig`.
         * `creationConfig` - configuration for queue and/or topic to create, if one does not exist. Should not be specified together with the `locatorConfig`.
+        * `messageTimestampField` - which field in the message contain the message creation date. This field needs to be ISO-8601 date string and by default it is `timestamp`, if your messages doesn't contain it the library will add one automatically to avoid infinite loops on consumer;
 * `init()`, prepare publisher for use (e. g. establish all necessary connections);
 * `close()`, stop publisher use (e. g. disconnect);
 * `publish()`, send a message to a queue or topic. It accepts the following parameters:
@@ -54,6 +55,8 @@ Multi-schema consumers support multiple message types via handler configs. They 
     * `options`, composed by
         * `handlers` – configuration for handling each of the supported message types. See "Multi-schema handler definition" for more details;
         * `messageTypeField` - which field in the message describes the type of a message. This field needs to be defined as `z.literal` in the schema and is used for routing the message to the correct handler;
+        * `messageTimestampField` - which field in the message contain the message creation date. This field needs to be ISO-8601 date string and by default it is `timestamp`;
+        * `maxRetryDuration` - how long (in seconds) the message should be retried due to the `retryLater` result before marking it as consumed (avoid infinite loops). Default is 4 days;
         * `queueName`; (for SNS publishers this is a misnomer which actually refers to a topic name)
         * `locatorConfig` - configuration for resolving existing queue and/or topic. Should not be specified together with the `creationConfig`.
         * `creationConfig` - configuration for queue and/or topic to create, if one does not exist. Should not be specified together with the `locatorConfig`.
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ They implement the following public methods:
         * `messageTypeField` - which field in the message describes the type of a message. This field needs to be defined as `z.literal` in the schema and is used for resolving the correct schema for validation
         * `locatorConfig` - configuration for resolving existing queue and/or topic. Should not be specified together with the `creationConfig`.
         * `creationConfig` - configuration for queue and/or topic to create, if one does not exist. Should not be specified together with the `locatorConfig`.
-        * `messageTimestampField` - which field in the message contain the message creation date. This field needs to be ISO-8601 date string and by default it is `timestamp`, if your messages doesn't contain it the library will add one automatically to avoid infinite loops on consumer;
+        * `messageTimestampField` - which field in the message contain the message creation date (by default it is `timestamp`). This field needs to be a `Date` object or ISO-8601 date string, if your messages doesn't contain it the library will add one automatically to avoid infinite loops on consumer;
 * `init()`, prepare publisher for use (e. g. establish all necessary connections);
 * `close()`, stop publisher use (e. g. disconnect);
 * `publish()`, send a message to a queue or topic. It accepts the following parameters:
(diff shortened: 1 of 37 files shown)


packages/sqs/lib/utils/sqsMessageDeserializer.ts
```diff
@@ -10,9 +10,10 @@ export const deserializeSQSMessage = <T extends object>(
   errorProcessor: ErrorResolver,
 ): Either<MessageInvalidFormatError | MessageValidationError, T> => {
   try {
-    return {
-      result: type.parse(JSON.parse(message.Body)),
-    }
+    const result = JSON.parse(message.Body)
+    type.parse(result)
+
+    return { result }
```
CarlosGamero, line 16: I didn't find a better way to do it, I tried to use `zod.passthrough()` but it is only available under `z.object` object but can't be applied to `ZodType` or `ZodSchema`

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -30,6 +31,7 @@ import { AbstractSqsService } from './AbstractSqsService'
 const ABORT_EARLY_EITHER: Either<'abort', never> = {
   error: 'abort',
 }
+const DEFAULT_MAX_RETRY_DURATION = 4 * 24 * 60 * 60 // 4 days in seconds
```
CarlosGamero, line 34: I think 4 days is the default time a message can be in the queue waiting

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -226,24 +228,35 @@ export abstract class AbstractSqsConsumer<
...
+                MessageBody: JSON.stringify({
+                  ...originalMessage,
+                  [this.messageTimestampField]: timestamp.toISOString(),
+                }),
+              }),
+            )
+          } else {
+            await this.failProcessing(message)
+          }
+
```
CarlosGamero, line 259: Decided to not split between barrier and handler `retryLater`, at the end in the handler is also possible to have a stuck message and we want to avoid that situation, let me know if you disagree

kibertoad: makes sense!

packages/amqp/lib/AbstractAmqpPublisher.ts
```diff
@@ -42,6 +42,14 @@ export abstract class AbstractAmqpPublisher<MessagePayloadType extends object>
     }
     resolveSchemaResult.result.parse(message)
 
+    // if the message doesn't have a timestamp field -> add it
```
kibertoad, line 45: might be useful to explain that this is needed to prevent infinite loops

packages/core/lib/utils/parseUtils.ts
```diff
@@ -3,15 +3,20 @@ import type { ZodSchema } from 'zod'
...
   errorProcessor: ErrorResolver,
-): Either<MessageInvalidFormatError | MessageValidationError, T> {
+): Either<MessageInvalidFormatError | MessageValidationError, ParseMessageResult<T>> {
   try {
-    return {
-      result: schema.parse(messagePayload),
-    }
+    const parsedMessage = schema.parse(messagePayload)
+
+    return { result: { parsedMessage, originalMessage: messagePayload as T } }
```
kibertoad, line 19: might be helpful to add comment, explaining why we have both

packages/amqp/lib/AbstractAmqpPublisher.ts
CarlosGamero: Will add it, thanks Igor 🙇

kibertoad (approved):

packages/amqp/lib/AbstractAmqpPublisher.ts
CarlosGamero: Also, I will update the doc at the end explaining the reason of the new parameters

packages/core/lib/utils/parseUtils.ts
CarlosGamero: Same as above

README.md
```diff
@@ -31,6 +31,7 @@ They implement the following public methods:
         * `messageTypeField` - which field in the message describes the type of a message. This field needs to be defined as `z.literal` in the schema and is used for resolving the correct schema for validation
         * `locatorConfig` - configuration for resolving existing queue and/or topic. Should not be specified together with the `creationConfig`.
         * `creationConfig` - configuration for queue and/or topic to create, if one does not exist. Should not be specified together with the `locatorConfig`.
+        * `messageTimestampField` - which field in the message contain the message creation date. This field needs to be ISO-8601 date string and by default it is `timestamp`, if your messages doesn't contain it the library will add one automatically to avoid infinite loops on consumer;
```
kibertoad, line 34: can we support Date too? nothing prevents zod schema from transforming timestamps within the schema

README.md
```diff
@@ -54,6 +55,8 @@ Multi-schema consumers support multiple message types via handler configs. They
     * `options`, composed by
         * `handlers` – configuration for handling each of the supported message types. See "Multi-schema handler definition" for more details;
         * `messageTypeField` - which field in the message describes the type of a message. This field needs to be defined as `z.literal` in the schema and is used for routing the message to the correct handler;
+        * `messageTimestampField` - which field in the message contain the message creation date. This field needs to be ISO-8601 date string and by default it is `timestamp`;
+        * `maxRetryDuration` - how long (in seconds) the message should be retried due to the `retryLater` result before marking it as consumed (avoid infinite loops). Default is 4 days;
```
kibertoad, line 59: we mark as consumed and not DLQ it?

packages/amqp/lib/AbstractAmqpConsumer.ts
```diff
@@ -128,37 +132,48 @@ export abstract class AbstractAmqpConsumer<
...
-          }
           if (result.result === 'success') {
             this.channel.ack(message)
-            this.handleMessageProcessed(deserializedMessage.result, 'consumed')
+            this.handleMessageProcessed(originalMessage, 'consumed')
+            return
+          }
+
+          // retryLater
+          const timestamp = this.tryToExtractTimestamp(originalMessage) ?? new Date()
```
kibertoad, line 161: maybe we should consider moving this logic to some util in core, as it's the same in all transports?

README.md
CarlosGamero: Yes, it goes to DLQ (only on SNS and AMQP) as any other message that can't be consumed, but only if DLQ is configured, already mentioned as a note on DLQ section 🙌 good catch Igor 🙏

packages/amqp/lib/AbstractAmqpConsumer.ts
CarlosGamero: Do you mean `this.tryToExtractTimestamp`? it is already added to `core` base class, if you mean `??` part I think that only for that doesn't make sense to add a new method but if you disagree I will more than happy to add it

kibertoad: Sorry for being ambiguous, I meant the date check/incrementation logic. adding dateUtils or something like that

CarlosGamero: Ah I see, yeah, will add it right now 🙇

README.md
CarlosGamero: Already done 🙇

packages/amqp/lib/AbstractAmqpConsumer.ts
CarlosGamero: done 🙌 thanks Igor 🙇

README.md
```diff
@@ -54,6 +55,8 @@ Multi-schema consumers support multiple message types via handler configs. They
     * `options`, composed by
         * `handlers` – configuration for handling each of the supported message types. See "Multi-schema handler definition" for more details;
         * `messageTypeField` - which field in the message describes the type of a message. This field needs to be defined as `z.literal` in the schema and is used for routing the message to the correct handler;
+        * `messageTimestampField` - which field in the message contain the message creation date (by default it is `timestamp`). This field needs to a `Date` object or ISO-8601 date string;
+        * `maxRetryDuration` - how long (in seconds) the message should be retried due to the `retryLater` result before marking it as consumed (avoid infinite loops). Default is 4 days;
```
kibertoad, line 59:
```suggestion
        * `maxRetryDuration` - how long (in seconds) the message should be retried due to the `retryLater` result before marking it as consumed (and sending to DLQ, if one is configured). This is used to avoid infinite loops. Default is 4 days;
```

packages/amqp/lib/AbstractAmqpConsumer.ts
```diff
@@ -128,37 +133,47 @@ export abstract class AbstractAmqpConsumer<
...
+          // requeue the message if maxRetryDuration is not exceeded, else ack it to avoid infinite loop
+          if (!isRetryDateExceeded(timestamp, this.maxRetryDuration)) {
+            this.channel.nack(message, false, true)
+            this.handleMessageProcessed(originalMessage, 'retryLater')
+          } else {
+            this.channel.ack(message)
+            this.handleMessageProcessed(originalMessage, 'error')
```
kibertoad, line 169: can we add a ToDo marker here to add DLQ invocation after we support it for RabbitMQ?

CarlosGamero: Good point! already added 🙌

kibertoad (approved):
Excellent job as always, Carlos!

README.md
```diff
@@ -31,6 +31,7 @@ They implement the following public methods:
         * `messageTypeField` - which field in the message describes the type of a message. This field needs to be defined as `z.literal` in the schema and is used for resolving the correct schema for validation
         * `locatorConfig` - configuration for resolving existing queue and/or topic. Should not be specified together with the `creationConfig`.
         * `creationConfig` - configuration for queue and/or topic to create, if one does not exist. Should not be specified together with the `locatorConfig`.
+        * `messageTimestampField` - which field in the message contain the message creation date (by default it is `timestamp`). This field needs to be a `Date` object or ISO-8601 date string, if your messages doesn't contain it the library will add one automatically to avoid infinite loops on consumer;
```
dariacm, line 34: 🟢 
```suggestion
        * `messageTimestampField` - which field in the message contains the message creation date (by default it is `timestamp`). This field needs to be a `Date` object or ISO-8601 date string, if your message doesn't contain it the library will add one automatically to avoid infinite loops on consumer;
```

README.md
```diff
@@ -54,6 +55,8 @@ Multi-schema consumers support multiple message types via handler configs. They
     * `options`, composed by
         * `handlers` – configuration for handling each of the supported message types. See "Multi-schema handler definition" for more details;
         * `messageTypeField` - which field in the message describes the type of a message. This field needs to be defined as `z.literal` in the schema and is used for routing the message to the correct handler;
+        * `messageTimestampField` - which field in the message contain the message creation date (by default it is `timestamp`). This field needs to a `Date` object or ISO-8601 date string;
```
dariacm, line 58: 🟢 
```suggestion
        * `messageTimestampField` - which field in the message contains the message creation date (by default it is `timestamp`). This field needs to be a `Date` object or an ISO-8601 date string;
```

dariacm (approved):

README.md
CarlosGamero: Thank you so much Daria!! 🧡

PR #142, opened 2024-05-17
Lazy init for AMQP

kibertoad:
Ensure that publisher initializes itself when invoked without active connection established.

### DIFF
--- a/packages/amqp/lib/AbstractAmqpPublisher.ts
+++ b/packages/amqp/lib/AbstractAmqpPublisher.ts
@@ -25,6 +25,8 @@ export abstract class AbstractAmqpPublisher<MessagePayloadType extends object>
 {
   private readonly messageSchemaContainer: MessageSchemaContainer<MessagePayloadType>
 
+  private initPromise?: Promise<void>
+
   constructor(dependencies: AMQPDependencies, options: AMQPPublisherOptions<MessagePayloadType>) {
     super(dependencies, options)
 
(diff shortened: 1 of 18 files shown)


packages/amqp/lib/AbstractAmqpPublisher.ts
```diff
@@ -42,6 +44,22 @@ export abstract class AbstractAmqpPublisher<MessagePayloadType extends object>
     }
     resolveSchemaResult.result.parse(message)
 
+    // If it's not initted yet, do the lazy init
+    if (!this.isInitted) {
+      // avoid multiple concurrent inits
+      if (!this.initPromise) {
+        this.initPromise = this.init()
+      }
+      this.initPromise
```
kibertoad, line 53: it is intentional that this promise is not awaited, that's how we keep the method invocation synchronous

RabbitMQ publish by itself doesn't guarantee that your message is delivered successfully, so this kind of fire-and-forget is not strongly different from how amqp-lib behaves in the first place.

kibertoad: moved to comment

packages/amqp/lib/AbstractAmqpPublisher.ts
```diff
@@ -42,6 +44,27 @@ export abstract class AbstractAmqpPublisher<MessagePayloadType extends object>
...
+       * RabbitMQ publish by itself doesn't guarantee that your message is delivered successfully, so this kind of fire-and-forget is not strongly different from how amqp-lib behaves in the first place.
+       */
+      this.initPromise
+        .then(() => {
+          this.publish(message)
+        })
+        .catch((err) => {
+          this.handleError(err)
+        })
+      return
```
CarlosGamero, line 65: 🟢 is `return` needed?

CarlosGamero (approved):
🚀

packages/amqp/lib/AbstractAmqpPublisher.ts
kibertoad: yes, otherwise we continue the synchronous execution and fail with an error, as there is no connection

CarlosGamero: Ah yeah, sorry I misread it 😓

PR #144, opened 2024-05-18
Implement AMQP publisherManager

kibertoad:


packages/amqp/vitest.config.mts
```diff
@@ -12,13 +12,21 @@ export default defineConfig({
...
     coverage: {
+      provider: 'v8',
       include: ['lib/**/*.ts'],
-      exclude: ['lib/**/*.spec.ts', 'lib/**/*.test.ts', 'test/**/*.*', 'lib/types/**/*.*'],
+      exclude: [
+        'lib/**/*.spec.ts',
+        'lib/**/*.test.ts',
+        'test/**/*.*',
+        'lib/types/**/*.*',
+        'lib/AmqpExchangePublisherManager.ts',
```
kibertoad, line 22: it's problematic to write tests for this functionality, as we don't yet have any logic in place for autocreating exchanges and autosubscribing queues. this will be done next.

packages/amqp/lib/AbstractAmqpQueuePublisher.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export type AmqpQueueMessageOptions = {
+  publishOptions: Options.Publish
+}
+
+const NO_PARAMS: AmqpQueueMessageOptions = {
+  publishOptions: {},
+}
+
+export class AbstractAmqpQueuePublisher<
```
kibertoad, line 13: This is a replacement for previously existing AbstractAmqpPublisher.

It's a breaking change.

packages/amqp/lib/AbstractAmqpExchangePublisher.ts
```diff
@@ -0,0 +1,50 @@
...
+> & {
+  exchange: string
+}
+
+export type AmqpExchangeMessageOptions = {
+  routingKey: string
+  publishOptions: Options.Publish
+}
+
+export class AbstractAmqpExchangePublisher<
```
kibertoad, line 22: This exposes RabbitMQ PubSub functionality (SNS-like).
previously we only had direct publishing to queues (SQS-like)

packages/amqp/lib/AbstractAmqpExchangePublisher.ts
```diff
@@ -0,0 +1,50 @@
...
+  MessagePayloadType extends object,
+> extends AbstractAmqpPublisher<MessagePayloadType, AmqpExchangeMessageOptions> {
+  constructor(
+    dependencies: AMQPDependencies,
+    options: AMQPExchangePublisherOptions<MessagePayloadType>,
+  ) {
+    super(dependencies, {
+      ...options,
+      // FixMe exchange publisher doesn't need queue at all
```
kibertoad, line 31: will be addressed during auto-init rework, as we need to implement exchange and subscription support

packages/core/lib/queues/AbstractPublisherManager.ts
```diff
@@ -1,25 +1,18 @@
...
-import type { SyncPublisher } from '../types/MessageQueueTypes'
-import type { CommonCreationConfigType } from '../types/queueOptionsTypes'
+import type { AsyncPublisher, SyncPublisher } from '../types/MessageQueueTypes'
+import type { CommonCreationConfigType, QueuePublisherOptions } from '../types/queueOptionsTypes'
 
-export type MessagePublishType<T extends CommonEventDefinition> = Pick<
-  z.infer<T['schema']>,
-  'type' | 'payload'
-> & { id?: string }
+export type MessagePublishType<T extends CommonEventDefinition> = z.infer<T['schema']>
```
kibertoad, line 10: Unfortunately, previous Pick-based implementation was breaking payload type inference during publishing. TS started accepting any combination of valid message types and payloads, and not enforcing that they belong to the same message

packages/sns/lib/sns/SnsPublisherManager.spec.ts
```diff
@@ -50,7 +43,13 @@ describe('SnsPublisherManager', () => {
       await fakeConsumer.start()
 
       // When
-      const publishedMessage = await publisherManager.publish(TestEvents.created.snsTopic, message)
+      const publishedMessage = await publisherManager.publish(TestEvents.created.snsTopic, {
+        ...publisherManager.resolveBaseFields(),
```
kibertoad, line 47: I wish we will be able to figure out a better way, that would both preserve specific payload type here, and not require explicitly passing all of the core fields, but so far I wasn't very successful in this.

packages/sns/lib/sns/SnsPublisherManager.spec.ts
```diff
@@ -80,13 +79,26 @@ describe('SnsPublisherManager', () => {
       await fakeConsumer.close()
     })
 
+    it('message publishing is type-safe', async () => {
+      await expect(
+        publisherManager.publish(TestEvents.created.snsTopic, {
+          ...publisherManager.resolveBaseFields(),
+          payload: {
+            // @ts-expect-error This should be causing a compilation error
```
kibertoad, line 87: this test can be used for checking the payload/type inference logic

packages/amqp/lib/AmqpExchangePublisherManager.ts
```diff
@@ -0,0 +1,96 @@
...
+import type { AMQPCreationConfig, AMQPDependencies, AMQPLocator } from './AbstractAmqpService'
+import type {
+  AmqpAwareEventDefinition,
+  AmqpMessageSchemaType,
+  AmqpPublisherManagerDependencies,
+  AmqpPublisherManagerOptions,
+} from './AmqpQueuePublisherManager'
+import { CommonAmqpExchangePublisherFactory } from './CommonAmqpPublisherFactory'
+
+import Omit = util.Omit
```
CarlosGamero, line 21: 🟢 nit: Importing omit could cause confusing with TS Omit, I would prefer using `util.Omit` or even `z.util.Omit` to make it more specific. But this is a very nitpicky comment, feel free to ignore

packages/amqp/lib/AmqpQueuePublisherManager.ts
```diff
@@ -0,0 +1,111 @@
...
+
+export type AmqpPublisherManagerDependencies<SupportedEvents extends AmqpAwareEventDefinition[]> = {
+  eventRegistry: EventRegistry<SupportedEvents>
+} & AMQPDependencies
+
+export type AmqpPublisherManagerOptions<
+  T extends AbstractAmqpPublisher<EventType, MO>,
+  MO,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  PO extends Omit<AMQPPublisherOptions<any>, 'creationConfig'>,
```
CarlosGamero, line 34: Not sure if possible, but just wondering if we could be possible to use unknown instead of any

CarlosGamero (approved):

packages/amqp/lib/AmqpExchangePublisherManager.ts
kibertoad: good catch, this was autoresolve by WebStorm, not intentional. I wanted to use TS Omit

packages/amqp/lib/AmqpQueuePublisherManager.ts
kibertoad: not unknown, but I'll try to wrestle with types to make this unnecessary

PR #145, opened 2024-05-19
Type-safe, automatically filled publisher metadata

kibertoad:
This separates publisher and consumer schemas; from publisher's perspective you may (but don't have to) fill id and timestamp, and from consumer's perspective they always receive them.

Includes logic for automatically generating both schemas.

This allows user to skip filling timestamp and id, while still strictly enforcing the payload structure


packages/amqp/lib/AmqpQueuePublisherManager.spec.ts
```diff
@@ -19,7 +19,6 @@ describe('AmqpQueuePublisherManager', () => {
       await fakeConsumer.start()
 
       const publishedMessage = await queuePublisherManager.publish(FakeConsumer.QUEUE_NAME, {
-        ...queuePublisherManager.resolveBaseFields(),
```
kibertoad, line 22: yes, this whole PR is to make this one line unnecessary :D

CarlosGamero (commented):
🙌😍🚀

CarlosGamero (approved):

PR #146, opened 2024-05-19
Separate AbstractAmqpConsumer from AbstractAmqpQueueConsumer

kibertoad:
Later we will add a different kind of an AbstractAmqpConsumer, AbstractAmqpExchangeConsumer (not final name, maybe AbstractAmqpTopicExchangeConsumer would be better)

See https://www.rabbitmq.com/tutorials/amqp-concepts


packages/amqp/lib/AmqpExchangePublisherManager.ts
```diff
@@ -37,7 +41,7 @@ export class AmqpExchangePublisherManager<
   >,
   SupportedEventDefinitions,
   MetadataType,
-  z.infer<SupportedEventDefinitions[number]['publisherSchema']>
+  AmqpExchangeMessageOptions
```
kibertoad, line 44: previous type was wrong

CarlosGamero (approved):

PR #150, opened 2024-05-23
AP-3648 SQS exponential delay between retries

CarlosGamero:


packages/amqp/lib/AbstractAmqpConsumer.ts
```diff
@@ -154,27 +154,26 @@ export abstract class AbstractAmqpConsumer<
         .then((result) => {
           if (result.result === 'success') {
             this.channel.ack(message)
-            this.handleMessageProcessed(originalMessage, 'consumed')
+            this.handleMessageProcessed(parsedMessage, 'consumed')
```
CarlosGamero, line 157: Changed to always use parsed message on spy as it can break tests using `toEqual`

packages/amqp/lib/AbstractAmqpConsumer.ts
```diff
@@ -154,27 +154,26 @@ export abstract class AbstractAmqpConsumer<
...
             return
           }
 
-          // retryLater
-          const timestamp = this.tryToExtractTimestamp(originalMessage) ?? new Date()
           // requeue the message if maxRetryDuration is not exceeded, else ack it to avoid infinite loop
-          if (!isRetryDateExceeded(timestamp, this.maxRetryDuration)) {
+          if (this.shouldBeRetried(originalMessage, this.maxRetryDuration)) {
+            // TODO: Add retry delay + republish message updating internal properties
```
CarlosGamero, line 163: Not implemented within this PR, adding todo

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -198,7 +203,45 @@ export abstract class AbstractQueueService<
...
+     */
+    const numberOfRetries = this.tryToExtractNumberOfRetries(message)
+    // @ts-ignore
+    messageCopy[this.messageNumberOfRetriesField] =
+      numberOfRetries !== undefined ? numberOfRetries + 1 : 0
+
+    return messageCopy
+  }
+
+  private tryToExtractTimestamp(message: MessagePayloadSchemas): Date | undefined {
```
CarlosGamero, line 244: Common logic here

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -42,11 +43,15 @@ export abstract class AbstractQueueService<
   ExecutionContext = undefined,
   PrehandlerOutput = undefined,
 > {
+  // Used to keep track of the number of retries performed on consumer
+  private readonly messageNumberOfRetriesField = '_internalNumberOfRetries'
```
kibertoad, line 47: `retry` naming sounds confusing for me in this context, because it's not clear how we are counting them - so first attempt is retry zero, and then second attempt is retry 1?

BullMQ naming of `attempts` is clearer for me; we can start at 1 (default when field is not set) and apply consistent counting logic.

packages/sqs/test/publishers/SqsPermissionPublisher.spec.ts
```diff
@@ -196,20 +224,32 @@ describe('SqsPermissionPublisher', () => {
 
       await permissionPublisher.publish(message)
 
-      const spy = await permissionPublisher.handlerSpy.waitForMessageWithId('1', 'published')
-      expect(spy.message).toEqual({ ...message, timestamp: expect.any(String) })
-      expect(spy.processingResult).toBe('published')
+      await waitAndRetry(() => !!receivedMessage)
```
kibertoad, line 227: can't we use `await myConsumer.handlerSpy.waitForMessage({
    // internal fields described here
})`
instead?

kibertoad: not sure if they are exposed to the handlerSpy, though

packages/core/lib/queues/AbstractQueueService.ts
CarlosGamero: Sounds good to me! will apply the change thanks Igor 🙇

packages/sqs/test/publishers/SqsPermissionPublisher.spec.ts
CarlosGamero: No, the number of retries is an internal property and it shouldn't be returned on the spy as it can break tests if someone is using `toEqual` (as we were doing on library tests)

kibertoad: gotcha! all good then

packages/core/lib/queues/AbstractQueueService.ts
CarlosGamero: Mmm, thinking about it, I am not really sure, this counting will only be applied on retryLater cases, meaning that errors will go with the SQS usual retry logic flow. 

`retry` here means, the number of times the consumer finished with `retryLater` result, naming it `attempt` I think can be more confusing, maybe something `_internalNumberOfRetryLater` to make it more clear?

kibertoad: fair point. maybe we can explain this with a jsdoc, without naming changes?

CarlosGamero: Of course! I will add it in a bit :D thanks Igor 🙏

kibertoad (approved):

PR #167, opened 2024-06-17
Large message payload offloading

eduard-fa:
This PR introduces a feature for offloading large message payloads to external storage.


packages/core/lib/messages/offloadedPayloadMessageSchemas.ts
```diff
@@ -0,0 +1,23 @@
+import { z } from 'zod'
+
+/**
+ * When the payload is too large to be sent in a single message, it is offloaded to a storage service and a pointer to the offloaded payload is sent instead.
+ * This schema represents the payload that is sent in place of the original payload.
+ */
+export const OFFLOADED_PAYLOAD_POINTER_PAYLOAD_SCHEME = z
```
kibertoad, line 7: SCHEME -> SCHEMA

packages/s3-payload-store/lib/S3PayloadStore.ts
```diff
@@ -0,0 +1,57 @@
...
+    } catch (e) {
+      if (e instanceof NoSuchKey) {
+        return null
+      }
+      throw e
+    }
+  }
+
+  async deletePayload(key: string) {
+    await this.s3.deleteObject({
```
kibertoad, line 52: Does this throw if object does not exist?

kibertoad: checked tests, seems like it doesn't

packages/core/lib/types/payloadStoreTypes.ts
```diff
@@ -0,0 +1,15 @@
+export interface PayloadStore {
+  /** Store the payload and return a key that can be used to retrieve it later. */
+  storePayload(payload: string): Promise<string>
```
kibertoad, line 3: would be nice to also support Readable, although in that case we also need to require passing the payload size as well

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -370,4 +375,76 @@ export abstract class AbstractQueueService<
...
+   */
+  protected async offloadMessagePayload(
+    message: MessagePayloadSchemas,
+    messageSize: number,
+  ): Promise<MessagePayloadSchemas | OffloadedPayloadPointerPayload> {
+    if (!this.payloadStoreConfig || messageSize <= this.payloadStoreConfig.messageSizeThreshold) {
+      return message
+    }
+
+    const payload = JSON.stringify(message)
```
kibertoad, line 392: Considering that we are dealing with large messages here, solution based on json-stream-stringify would be nicer (we could utilize https://github.com/lokalise/node-core/blob/main/src/utils/streamUtils.ts) here, but this can be done as a separate optimization later

packages/sqs/lib/sqs/AbstractSqsPublisher.ts
```diff
@@ -64,15 +72,12 @@ export abstract class AbstractSqsPublisher<MessagePayloadType extends object>
       }
 
       message = this.updateInternalProperties(message)
+      const maybeOffloadedPayloadMessage = await this.offloadMessagePayload(
+        message,
+        calculateOutgoingMessageSize(message),
```
kibertoad, line 77: can we do the calculation inside, after we have already passed the check of `is offloading enabled`? otherwise we are making the flow slower for everyone, including those who don't use the functionality

packages/sqs/lib/sqs/AbstractSqsPublisher.ts
```diff
@@ -64,15 +72,12 @@ export abstract class AbstractSqsPublisher<MessagePayloadType extends object>
       }
 
       message = this.updateInternalProperties(message)
+      const maybeOffloadedPayloadMessage = await this.offloadMessagePayload(
```
kibertoad, line 75: `offloadMessagePayloadIfNeeded` might be a more accurate name, as it is conditional

packages/sqs/lib/utils/sqsUtils.ts
```diff
@@ -219,3 +224,12 @@ export async function deleteQueue(
     console.log(`Failed to delete: ${err.message}`)
   }
 }
+
+export function calculateOutgoingMessageSize(message: unknown) {
+  const messageBody = JSON.stringify(message)
+  return calculateSqsMessageBodySize(messageBody)
+}
+
+export function calculateSqsMessageBodySize(messageBody: SendMessageCommandInput['MessageBody']) {
```
kibertoad, line 233: what's SQS-specific to this?

packages/core/lib/messages/offloadedPayloadMessageSchemas.ts
eduard-fa: Fixed, thanks

packages/sqs/lib/utils/sqsUtils.ts
eduard-fa: Only the SQS layer is aware of how the message body is represented (serialized) in the message and what kind of extra attributes may be present that will impact the resulting message size.

packages/sqs/lib/sqs/AbstractSqsPublisher.ts
eduard-fa: Fair point.
I was trying to avoid requiring calculateOutgoingMessageSize() in the abstraction and then stubbing it in the Consumer implementation.
What if we pass the size calculation as a callable that will be invoked when needed?
Screenshot 2024-06-17 at 15 22 58

packages/core/lib/queues/AbstractQueueService.ts
eduard-fa: Implemented, but I'm not fully convinced that we want to use stream-stringify all the time due to the filesystem overhead. It might be that the final solution should either do everything in memory or allow more granular control over when to dump messages to the filesystem.

packages/core/lib/types/payloadStoreTypes.ts
eduard-fa: Considering that the payload store is used internally, it does not make much sense to support both types, so I've changed it to be just readable.

packages/core/lib/utils/streamUtils.ts
```diff
@@ -0,0 +1,18 @@
+import type { Readable } from 'node:stream'
+
+export async function streamWithKnownSizeToString(stream: Readable, size: number): Promise<string> {
```
kibertoad, line 3: let's move it to node-core later, but works for now

packages/s3-payload-store/package.json
```diff
@@ -0,0 +1,68 @@
+{
+  "name": "@message-queue-toolkit/s3-store",
+  "version": "0.1.0",
+  "private": false,
+  "license": "MIT",
+  "description": "AWS S3-based message store implementation for message-queue-toolkit",
+  "maintainers": [
+    {
+      "name": "Igor Savin",
```
kibertoad, line 9: shall we add @eduard-fa here?

packages/sqs/lib/sqs/AbstractSqsPublisher.ts
kibertoad: yup, sounds good!

packages/core/lib/types/payloadStoreTypes.ts
kibertoad: how will that work in systems that don't support writing to FS? we might need some fallback there

packages/sqs/test/consumers/SqsPermissionConsumer.spec.ts
```diff
@@ -485,6 +487,60 @@ describe('SqsPermissionConsumer', () => {
...
+        },
+      } satisfies PERMISSIONS_ADD_MESSAGE_TYPE
+
+      await publisher.publish(message)
+
+      const consumptionResult = await consumer.handlerSpy.waitForMessageWithId(
+        message.id,
+        'consumed',
+      )
+      expect(consumptionResult.message).toMatchObject(message)
```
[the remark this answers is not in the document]
eduard-fa: Well, we are explicitly using a message that is larger than the message size limit: https://github.com/kibertoad/message-queue-toolkit/blob/f0206c5528fdcbb6f6d03d538a5268f68efefca3/packages/sqs/test/consumers/SqsPermissionConsumer.spec.ts#L491

and it indeed fails if payload offloading isn't enabled.

packages/core/lib/queues/AbstractQueueService.ts
kibertoad: I agree, it needs to be configureable in some shape. Some users may not be able to use FS at all

packages/sqs/lib/utils/sqsUtils.ts
kibertoad: can you add a comment with explanation how it will need to be changed in case we'll start using attributes there?

packages/core/lib/types/payloadStoreTypes.ts
eduard-fa: Please check this implementation: https://github.com/kibertoad/message-queue-toolkit/pull/167/commits/1fb4c534068227213de5d855442eab802a710c0a. It should provide more flexibility, including avoiding file system-based serialization.

packages/core/lib/queues/AbstractQueueService.ts
eduard-fa: Made changes as part of https://github.com/kibertoad/message-queue-toolkit/pull/167#discussion_r1643916743. The default behavior is still based on `JsonStreamStringify` + `FsReadableProvider`, but now there is a possibility to override the serialization logic.

packages/s3-payload-store/package.json
eduard-fa: If you don't mind

README.md
```diff
@@ -242,3 +244,81 @@ const result = await myConsumer.handlerSpy.waitForMessageWithId('1')
...
+    
+    const payloadStoreConfig = {
+        store: new S3PayloadStore(
+            {s3: s3Client},
+            {
+                bucketName: 'your-s3-bucket-name',
+                keyPrefix: 'optional-key-prefix'
+            }
+        ),
+        messageSizeThreshold: 256 * 1024 // 256KB
```
kibertoad, line 279: Can we expose limits, compatible with SNS and SQS as constants from according packages?

packages/core/lib/payload-store/JsonStringifyStreamSerializer.ts
```diff
@@ -0,0 +1,27 @@
...
+export type TemporaryFilePathResolver = () => string
+const defaultTemporaryFilePathResolver: TemporaryFilePathResolver = () => tmpNameSync()
+
+export class JsonStringifyStreamSerializer implements PayloadSerializer {
+  constructor(
+    private readonly temporaryFilePathResolver: TemporaryFilePathResolver = defaultTemporaryFilePathResolver,
+  ) {}
+
+  async serialize(payload: unknown) {
+    const fsReadableProvider = await FsReadableProvider.persistReadableToFs({
```
kibertoad, line 16: this is going to leak files. serializer interface should also expose close/cleanup method, which in this case should destroy the provider

README.md
eduard-fa: Done: https://github.com/kibertoad/message-queue-toolkit/pull/167/commits/fff4268f892dedb12da86e5aa673295c5f041648

packages/s3-payload-store/package.json
kibertoad: please do!

README.md
```diff
@@ -242,3 +244,81 @@ const result = await myConsumer.handlerSpy.waitForMessageWithId('1')
...
+                payloadStoreConfig,
+            })
+        }
+    }
+    ```
+3. **Usage:**
+   Messages larger than the `messageSizeThreshold` are automatically offloaded to the configured store (S3 bucket in example). 
+   The queue manages only references to these payloads, ensuring compliance with size limits.
+4. **Managing payload lifecycle:**
+   It is important to note that once messages are processed, their payloads are not automatically deleted from the external storage.
```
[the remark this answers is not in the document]
eduard-fa: The primary reason is simplification. 
Managing payloads becomes complicated with fan-out, as we need to track if all consumers have read the message, which is a challenging task. 

Additionally, you'll likely need to set up some expiration policy for unconsumed or DLQ-rerouted messages since you can't keep them indefinitely. This leads to having an expiration policy in place anyway.

packages/core/lib/payload-store/JsonStringifyStreamSerializer.ts
eduard-fa: Addressed in https://github.com/kibertoad/message-queue-toolkit/pull/167/commits/bbe6f6ae8634e5ec65037a7390ed077918b83718

README.md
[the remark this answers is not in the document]
kibertoad: Fair point. Maybe we should mention that in the documentation as well?

packages/sns/lib/sns/AbstractSnsPublisher.ts
```diff
@@ -118,5 +121,29 @@ export abstract class AbstractSnsPublisher<MessagePayloadType extends object>
...
     throw new Error('Not implemented for publisher')
   }
+
+  protected async sendMessage(
+    payload: MessagePayloadType | OffloadedPayloadPointerPayload,
+    options: SNSMessageOptions,
+  ): Promise<void> {
+    const attributes: Record<string, MessageAttributeValue> = {}
+
+    if (isOffloadedPayloadPointerPayload(payload)) {
```
kibertoad, line 131: this looks suspiciously identical to SQS implementation, maybe we should move it to some util?

README.md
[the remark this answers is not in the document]
eduard-fa: Expanded section with the extra note: https://github.com/kibertoad/message-queue-toolkit/pull/167/commits/d3224e9fb3812cfd4a6ca230955b15c43d0d9579

kibertoad (approved):

packages/sns/lib/sns/AbstractSnsPublisher.ts
eduard-fa: Fair point
Changed: https://github.com/kibertoad/message-queue-toolkit/pull/167/commits/1e45a4020d0d3e28cfef275ca66d40ac18fca217

PR #174, opened 2024-06-20
Add PublisherManager documentation to README

dariacm:


README.md
```diff
@@ -44,6 +44,75 @@ They implement the following public methods:
...
+          messageTypeField: 'type',
+          creationConfig: {
+            updateAttributesIfExists: true,
+          },
+        },
+      })
+    },
+    {
+      lifetime: Lifetime.SINGLETON,
+      enabled: queuesEnabled,
```
kibertoad, line 74: queuesEnabled is unclear in this context

README.md
```diff
@@ -44,6 +44,75 @@ They implement the following public methods:
...
 > **_NOTE:_**  See [SqsPermissionPublisher.ts](./packages/sqs/test/publishers/SqsPermissionPublisher.ts) for a practical example.
 
+#### PublisherManager
+
+PublisherManager is a wrapper to automatically spawn publishers on demand and reduce the amount of boilerplate needed to write them, while guaranteeing type-safety and autocompletion. It also automatically fills metadata fields.
+
+You can create an instance of PublisherManager as shown in this example:
+
+```typescript
+const diConfig: DiConfig = {
```
kibertoad, line 54: I would remove DiConfig from the example, users of the library may not be using our template or awilix altogether

README.md
```diff
@@ -44,6 +44,75 @@ They implement the following public methods:
...
+  * `message` – the message to be published;
+  * `precedingEventMetadata` – (optional) ;
+  * `messageOptions` – (optional) a protocol-dependent set of message parameters. For more information please check documentation for options for [AMQP](https://amqp-node.github.io/amqplib/channel_api.html#channel_sendToQueue).
+* `publishSync()` (AMQP), sends a message to a queue or topic. Attempts to publish to  It accepts the following parameters:
+  * `queue` or `topic` – the queue or topic to publish to;
+  * `message` – the message to be published;
+  * `precedingEventMetadata` – (optional) metadata that describes the context in which the message was created. If not provided, each field will resolve to its default value. Available fields are:
+    * `schemaVersion` – message schema version, defaults to 1.0.0;
+    * `producedBy` – service that produced the message, defaults to the service ID provided in `CommonMetadataFiller`;
+    * `originatedFrom` – service that initiated the entire workflow that led to this message, defaults to the service ID provided in `CommonMetadataFiller`;
```
kibertoad, line 109: it also reuses value from precedingEventMetadata if set

README.md
```diff
@@ -44,6 +44,75 @@ They implement the following public methods:
...
+  * `topic` – the topic to publish to;
+  * `message` – the message to be published;
+  * `precedingEventMetadata` – (optional) ;
+  * `messageOptions` – (optional) a protocol-dependent set of message parameters. For more information please check documentation for options for [AMQP](https://amqp-node.github.io/amqplib/channel_api.html#channel_sendToQueue).
+* `publishSync()` (AMQP), sends a message to a queue or topic. Attempts to publish to  It accepts the following parameters:
+  * `queue` or `topic` – the queue or topic to publish to;
+  * `message` – the message to be published;
+  * `precedingEventMetadata` – (optional) metadata that describes the context in which the message was created. If not provided, each field will resolve to its default value. Available fields are:
+    * `schemaVersion` – message schema version, defaults to 1.0.0;
+    * `producedBy` – service that produced the message, defaults to the service ID provided in `CommonMetadataFiller`;
```
kibertoad, line 108: bttw, this should actually never take value from precedingMetadata, as it it is always equal to the service where metadatafiller is located. if we want to override that, we need a separate field for that

README.md
dariacm: > metadata that describes the context in which the message was created. **If not provided, each field will resolve to its default value.** 

Is this not clear enough? I am trying to say that it uses what's provided in `precedingEventMetadata` if any and if not provided will fallback on the default value.

README.md
dariacm: Oh I am confused 🙈 was https://github.com/kibertoad/message-queue-toolkit/pull/169 wrong then?

kibertoad: a little bit. my mistake, forgot the original intent behind it

dariacm: I've removed it from here and will open another PR to restore its behavior then

kibertoad (approved):

PR #184, opened 2024-07-04
Support synchronous check for whether the message was published

kibertoad:
This is primarily useful for checking that message _wasn't_ published.

### DIFF
--- a/packages/sns/package.json
+++ b/packages/sns/package.json
@@ -25,7 +25,7 @@
     "prepublishOnly": "npm run build:release"
   },
   "dependencies": {
-    "@lokalise/node-core": "^10.0.0",
+    "@lokalise/node-core": "^10.0.1",
     "sqs-consumer": "^10.3.0",
     "zod": "^3.23.8"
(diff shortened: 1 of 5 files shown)


packages/sns/package.json
```diff
@@ -25,7 +25,7 @@
     "prepublishOnly": "npm run build:release"
   },
   "dependencies": {
-    "@lokalise/node-core": "^10.0.0",
+    "@lokalise/node-core": "^10.0.1",
```
CarlosGamero, line 28: Should we update it also on AMQP and SQS?

CarlosGamero (approved):

packages/sns/package.json
kibertoad: they are already up-to-date, as far as I can see

CarlosGamero: Oh, sorry 🙇, thank youu 🙏

kibertoad: np!

PR #201, opened 2024-08-28
AP-4870 async event handlers support

CarlosGamero:


packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -79,52 +98,19 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
-      })
-    }
-
-    const eventHandlers = this.eventHandlerMap[eventTypeName]
-
-    // No relevant handlers are registered, we can stop processing
-    if (!eventHandlers && this.anyHandlers.length === 0) {
-      // @ts-ignore
-      return data
-    }
```
CarlosGamero, line 102: Removing this early return because we might not have handlers on `eventHandlerMap`, but we can have them on `anyHandlers`

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -16,36 +17,50 @@ import type {
   SingleEventHandler,
 } from './eventTypes'
 
+export type DomainEventEmitterDependencies<SupportedEvents extends CommonEventDefinition[]> = {
+  eventRegistry: EventRegistry<SupportedEvents>
+  metadataFiller: MetadataFiller
+  // TODO: make them mandatory is a breaking change, decide if we are fine with that
```
CarlosGamero, line 23: TBD

CarlosGamero: After thinking about it, I think that we make logger and error-reporter mandatory, as it is really useful to understand if a BG listener failed

packages/schemas/lib/events/eventTypes.ts
```diff
@@ -39,6 +43,7 @@ export type EventHandler<
   EventDefinitionSchema extends
     CommonEventDefinitionPublisherSchemaType<CommonEventDefinition> = CommonEventDefinitionPublisherSchemaType<CommonEventDefinition>,
 > = {
+  readonly eventHandlerId: string
   handleEvent(event: EventDefinitionSchema): void | Promise<void>
 }
```
CarlosGamero, line 48: We need to add this to be able to add observability on bg handlers

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -151,27 +150,69 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
+    const bgHandlers = [...eventHandlers.background, ...this.anyHandlers.background]
+    for (const handler of bgHandlers) {
+      const transactionId = randomUUID()
+      // not sure if we should use startWithGroup or start, using group to group all handlers for the same event type
+      // should it be eventId + eventType or just eventType?
+      this.transactionObservabilityManager?.startWithGroup(
+        this.buildTransactionKey(event, handler),
+        transactionId,
+        event.type,
+      )
```
CarlosGamero, line 190: Please share your thoughts 🙏! If you agree I will remove the comment

packages/schemas/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@message-queue-toolkit/schemas",
-  "version": "3.0.0",
+  "version": "4.0.0",
```
kibertoad, line 3: why bump semver major?

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -151,27 +150,69 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
+      await handler.handleEvent(event)
     }
 
-    this.eventHandlerMap[eventTypeName].push(handler)
+    const bgHandlers = [...eventHandlers.background, ...this.anyHandlers.background]
+    for (const handler of bgHandlers) {
+      const transactionId = randomUUID()
+      // not sure if we should use startWithGroup or start, using group to group all handlers for the same event type
+      // should it be eventId + eventType or just eventType?
+      this.transactionObservabilityManager?.startWithGroup(
```
kibertoad, line 186: we don't need the same thing for non-background handlers?

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -141,8 +134,14 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
   public on<EventTypeName extends EventTypeNames<SupportedEvents[number]>>(
     eventTypeName: EventTypeName,
     handler: SingleEventHandler<SupportedEvents, EventTypeName>,
+    bgHandler = false,
```
kibertoad, line 137: maybe `isBackgroundHandler`?

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -16,36 +23,52 @@ import type {
   SingleEventHandler,
 } from './eventTypes'
 
+export type DomainEventEmitterDependencies<SupportedEvents extends CommonEventDefinition[]> = {
+  eventRegistry: EventRegistry<SupportedEvents>
+  metadataFiller: MetadataFiller
+  logger: Logger
+  errorReporter: ErrorReporter
```
kibertoad, line 30: we should provide some dummy reporter implementation that just logs errors in case we want to make this mandatory

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: Do we? it was not there before but I think it would be beneficial so will add it

packages/schemas/package.json
CarlosGamero: The `handler` type ha a new required variable that must be filled, it implies a breaking change

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: Good point! will do it 🙏

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: Mmm, we already have the logger there as mandatory and is doing exactly that, so I would say we can make it option

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: Done 🙌, thanks Igor for raising it 🙇

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: Done, thank youuu

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: Change applied, thank youuu 🙏

packages/schemas/package.json
kibertoad: but why does that affect schemas library? nothing changed for it

packages/core/lib/events/DomainEventEmitter.ts
kibertoad: we used to check both in this step. it's no longer possible?

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -151,27 +150,82 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
-      this.eventHandlerMap[eventTypeName] = []
+  private async handleEvent<SupportedEvent extends SupportedEvents[number]>(
+    event: CommonEventDefinitionPublisherSchemaType<SupportedEvent>,
+  ): Promise<void> {
+    const eventHandlers = this.eventHandlerMap[event.type] ?? {
+      foreground: [],
+      background: [],
+    }
+
+    const fgHandlers = [...eventHandlers.foreground, ...this.anyHandlers.foreground]
```
kibertoad, line 176: it seems a bit wasteful to repopulate this on every event. can we do that after handlers are attached instead?

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -151,27 +150,82 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
+    for (const handler of fgHandlers) {
+      const transactionId = randomUUID()
+      let isSuccessfull = false
+      try {
+        this.transactionObservabilityManager?.startWithGroup(
```
kibertoad, line 181: yeah, I'd say it makes sense to use startWithGroup here

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: Discussed on Slack, for visibility:
To avoid creating this array with every event, we moved any handler to `eventHandlerMap`

CarlosGamero: Already done :D

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: You are right! we were checking both, I can put the early return back if you think it makes sense, but I am not sure if it is really useful.
From my point of view, no matter if there are no handlers assigned to the event, event should validated (to avoid errors on old events if we attach a new handler) and spy should have it into account that is what we are doing below, what do you think?

kibertoad: fair point!

kibertoad (approved):

PR #204, opened 2024-09-04
AP-5046 outbox-core package for transactional outbox pattern

kamilwylegala:
Added few classes in outbox.ts that cover:
- Periodic job that runs every 1 sec to process entries from outbox storage
- OutboxStorage interface -> next step, implement prisma

I didn't decide to implement `lockedUntil` for `ACKed` entries. Now the code that handles entries is wrapped with try-catch, in case of errors we always end up with `FAILED` status, next execution will pick up such entries until retry count max is reached.

### DIFF
--- a/packages/outbox-core/README.md
+++ b/packages/outbox-core/README.md
@@ -35,8 +35,29 @@ const job = new OutboxPeriodicJob(
 
 Job will take care of processing outbox entries emitted by:
 ```typescript
-const emitter = new OutboxEventEmitter(
+import {
+  type CommonEventDefinition,
+  enrichMessageSchemaWithBase,
+} from '@message-queue-toolkit/schemas'
+
+const MyEvents = {
+  created: {
+    ...enrichMessageSchemaWithBase(
+      'entity.created',
+      z.object({
+        message: z.string(),
+      }),
+    ),
+  },
+} as const satisfies Record<string, CommonEventDefinition>
+
+type MySupportedEvents = (typeof TestEvents)[keyof typeof TestEvents][]
+
+const emitter = new OutboxEventEmitter<MySupportedEvents>(
     //Same instance of outbox storage that is used by OutboxPeriodicJob
     outboxStorage
 )
+
+//It pushes the entry to the storage, later will be picked up by the OutboxPeriodicJob
+await emitter.emit(/* args */)
 ```
--- a/packages/outbox-core/README.md
+++ b/packages/outbox-core/README.md
@@ -5,7 +5,7 @@ Main package that contains the core functionality of the Outbox pattern to provi
 ## Installation
 
 ```bash
-npm i -S outbox-core
+npm i -S @message-queue-toolkit/outbox-core
 ```
 
 ## Usage
--- a/packages/outbox-core/README.md
+++ b/packages/outbox-core/README.md
@@ -1,3 +1,42 @@
 # outbox-core
 
-WIP
+Main package that contains the core functionality of the Outbox pattern to provide "at least once" delivery semantics for messages.
+
+## Installation
+
+```bash
+npm i -S outbox-core
+```
+
+## Usage
+
+To process outbox entries and emit them to the message queue, you need to create an instance of the `OutboxPeriodicJob` class:
+
+```typescript
+import { OutboxPeriodicJob } from '@message-queue-toolkit/outbox-core';
+
+const job = new OutboxPeriodicJob(
+    //Implementation of OutboxStorage interface, TODO: Point to other packages in message-queue-toolkit
+    outboxStorage, 
+    //Default available accumulator for gathering outbox entries as the process job is progressing.
+    new InMemoryOutboxAccumulator(), //Default accumulator
+    //DomainEventEmitter, it will be used to publish events, see @message-queue-toolkit/core
+    eventEmitter,
+    //See PeriodicJobDependencies from @lokalise/background-jobs-common
+    dependencies,
+    //Retry count, how many times outbox entries should be retried to be processed
+    3,
+    //emitBatchSize - how many outbox entries should be emitted at once
+    10,
+    //internalInMs - how often the job should be executed, e.g. below it runs every 1sec
+    1000
+)
+```
+
+Job will take care of processing outbox entries emitted by:
+```typescript
+const emitter = new OutboxEventEmitter(
+    //Same instance of outbox storage that is used by OutboxPeriodicJob
+    outboxStorage
+)
+```
--- a/packages/outbox-core/README.md
+++ b/packages/outbox-core/README.md
@@ -19,7 +19,7 @@ const job = new OutboxPeriodicJob(
     //Implementation of OutboxStorage interface, TODO: Point to other packages in message-queue-toolkit
     outboxStorage, 
     //Default available accumulator for gathering outbox entries as the process job is progressing.
-    new InMemoryOutboxAccumulator(), //Default accumulator
+    new InMemoryOutboxAccumulator(),
     //DomainEventEmitter, it will be used to publish events, see @message-queue-toolkit/core
     eventEmitter,
     //See PeriodicJobDependencies from @lokalise/background-jobs-common
--- a/packages/outbox-core/README.md
+++ b/packages/outbox-core/README.md
@@ -1,0 +1,3 @@
+# outbox-core
+
+WIP
--- a/packages/outbox-core/lib/accumulators.ts
+++ b/packages/outbox-core/lib/accumulators.ts
@@ -1,15 +1,41 @@
 import type { CommonEventDefinition } from '@message-queue-toolkit/schemas'
 import type { OutboxEntry } from './objects.ts'
 
+/**
+ * Accumulator is responsible for storing outbox entries in two cases:
+ * - successfully dispatched event
+ * - failed events
+ *
+ * Thanks to this, we can use aggregated result and persist in the storage in batches.
+ */
 export interface OutboxAccumulator<SupportedEvents extends CommonEventDefinition[]> {
+  /**
+   * Accumulates successfully dispatched event.
+   * @param outboxEntry
+   */
   add(outboxEntry: OutboxEntry<SupportedEvents[number]>): Promise<void>
 
+  /**
+   * Accumulates failed event.
+   * @param outboxEntry
+   */
   addFailure(outboxEntry: OutboxEntry<SupportedEvents[number]>): Promise<void>
 
+  /**
+   * It's meant to be used by OutboxStorage::flush() to get all entries that should be persisted as successful ones.
+   */
   getEntries(): Promise<OutboxEntry<SupportedEvents[number]>[]>
 
+  /**
+   * Also used by OutboxStorage::flush() to get all entries that should be persisted as failed ones. Such entries will be retried + their retryCount will be incremented.
+   */
   getFailedEntries(): Promise<OutboxEntry<SupportedEvents[number]>[]>
 
+  /**
+   * After running clear(), no entries should be returned by getEntries() and getFailedEntries().
+   *
+   * clear() is always called after flush() in OutboxStorage.
+   */
   clear(): Promise<void>
 }
 
--- a/packages/outbox-core/lib/accumulators.ts
+++ b/packages/outbox-core/lib/accumulators.ts
@@ -22,12 +22,12 @@ export interface OutboxAccumulator<SupportedEvents extends CommonEventDefinition
   addFailure(outboxEntry: OutboxEntry<SupportedEvents[number]>): Promise<void>
 
   /**
-   * It's meant to be used by OutboxStorage::flush() to get all entries that should be persisted as successful ones.
+   * Returns all entries that should be persisted as successful ones.
    */
   getEntries(): Promise<OutboxEntry<SupportedEvents[number]>[]>
 
   /**
-   * Also used by OutboxStorage::flush() to get all entries that should be persisted as failed ones. Such entries will be retried + their retryCount will be incremented.
+   * Returns all entries that should be persisted as failed ones. Such entries will be retried + their retryCount will be incremented.
    */
   getFailedEntries(): Promise<OutboxEntry<SupportedEvents[number]>[]>
 
--- a/packages/outbox-core/lib/accumulators.ts
+++ b/packages/outbox-core/lib/accumulators.ts
@@ -1,0 +1,47 @@
+import type { CommonEventDefinition } from '@message-queue-toolkit/schemas'
+import type { OutboxEntry } from './objects.ts'
+
+export interface OutboxAccumulator<SupportedEvents extends CommonEventDefinition[]> {
+  add(outboxEntry: OutboxEntry<SupportedEvents[number]>): Promise<void>
+
+  addFailure(outboxEntry: OutboxEntry<SupportedEvents[number]>): Promise<void>
+
+  getEntries(): Promise<OutboxEntry<SupportedEvents[number]>[]>
+
+  getFailedEntries(): Promise<OutboxEntry<SupportedEvents[number]>[]>
+
+  clear(): Promise<void>
+}
+
+export class InMemoryOutboxAccumulator<SupportedEvents extends CommonEventDefinition[]>
+  implements OutboxAccumulator<SupportedEvents>
+{
+  private entries: OutboxEntry<SupportedEvents[number]>[] = []
+  private failedEntries: OutboxEntry<SupportedEvents[number]>[] = []
+
+  public add(outboxEntry: OutboxEntry<SupportedEvents[number]>) {
+    this.entries = [...this.entries, outboxEntry]
+
+    return Promise.resolve()
+  }
+
+  public addFailure(outboxEntry: OutboxEntry<SupportedEvents[number]>) {
+    this.failedEntries = [...this.failedEntries, outboxEntry]
+
+    return Promise.resolve()
+  }
+
+  getEntries(): Promise<OutboxEntry<SupportedEvents[number]>[]> {
+    return Promise.resolve(this.entries)
+  }
+
+  getFailedEntries(): Promise<OutboxEntry<SupportedEvents[number]>[]> {
+    return Promise.resolve(this.failedEntries)
+  }
+
+  public clear(): Promise<void> {
+    this.entries = []
+    this.failedEntries = []
+    return Promise.resolve()
+  }
+}
--- a/packages/outbox-core/lib/accumulators.ts
+++ b/packages/outbox-core/lib/accumulators.ts
@@ -46,13 +46,13 @@ export class InMemoryOutboxAccumulator<SupportedEvents extends CommonEventDefini
   private failedEntries: OutboxEntry<SupportedEvents[number]>[] = []
 
   public add(outboxEntry: OutboxEntry<SupportedEvents[number]>) {
-    this.entries = [...this.entries, outboxEntry]
+    this.entries.push(outboxEntry)
 
     return Promise.resolve()
   }
 
   public addFailure(outboxEntry: OutboxEntry<SupportedEvents[number]>) {
-    this.failedEntries = [...this.failedEntries, outboxEntry]
+    this.failedEntries.push(outboxEntry)
 
     return Promise.resolve()
   }
--- a/packages/outbox-core/lib/outbox.spec.ts
+++ b/packages/outbox-core/lib/outbox.spec.ts
@@ -49,31 +49,6 @@ const createdEventPayload: CommonEventDefinitionPublisherSchemaType<typeof TestE
   },
 }
 
-const updatedEventPayload: CommonEventDefinitionPublisherSchemaType<typeof TestEvents.updated> = {
-  ...createdEventPayload,
-  type: 'entity.updated',
-}
-
-const expectedCreatedPayload = {
-  id: expect.any(String),
-  timestamp: expect.any(String),
-  payload: {
-    message: 'msg',
-  },
-  type: 'entity.created',
-  metadata: {
-    correlationId: expect.any(String),
-    originatedFrom: 'service',
-    producedBy: 'producer',
-    schemaVersion: '1',
-  },
-}
-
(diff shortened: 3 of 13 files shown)


kamilwylegala:
@kibertoad Ready for review.

packages/outbox-core/lib/outbox.ts
```diff
@@ -0,0 +1,170 @@
...
+    outboxStorage: OutboxStorage<SupportedEvents>,
+    eventEmitter: DomainEventEmitter<SupportedEvents>,
+    maxRetryCount: number,
+    dependencies: PeriodicJobDependencies,
+  ) {
+    super(
+      {
+        jobId: 'OutboxJob',
+        schedule: {
+          intervalInMs: 1000,
```
kibertoad, line 119: Would be nice to make this configureable

packages/outbox-core/lib/outbox.ts
```diff
@@ -0,0 +1,170 @@
+import { AbstractPeriodicJob, type JobExecutionContext } from '@lokalise/background-jobs-common'
+import type { PeriodicJobDependencies } from '@lokalise/background-jobs-common/dist/periodic-jobs/periodicJobTypes'
```
kibertoad, line 2: if we are missing some types there, it's preferable to adjust exports on the packages rather than use the dist import

packages/outbox-core/lib/outbox.ts
```diff
@@ -0,0 +1,170 @@
...
+import { generateUuid7 } from '@lokalise/id-utils'
```
kibertoad, line 3: considering that this is a public library, I'd suggest including uuid7 library as a direct dependency instead

packages/outbox-core/lib/outbox.ts
```diff
@@ -0,0 +1,170 @@
...
+    private readonly eventEmitter: DomainEventEmitter<SupportedEvents>,
+    private readonly maxRetryCount: number,
+  ) {}
+
+  public async processOutboxEntries(context: JobExecutionContext) {
+    const entries = await this.outboxStorage.getEntries(this.maxRetryCount)
+
+    for (const entry of entries) {
+      try {
+        const updatedEntry = await this.outboxStorage.update({
```
kibertoad, line 74: what if my service instance dies after acking but before the emit happens?

packages/outbox-core/lib/outbox.ts
```diff
@@ -0,0 +1,170 @@
...
+          ...entry,
+          updated: new Date(),
+          status: 'ACKED',
+        })
+
+        await this.eventEmitter.emit(entry.event, entry.data, entry.precedingMessageMetadata)
+
+        await this.outboxStorage.update({ ...updatedEntry, updated: new Date(), status: 'SUCCESS' })
```
kibertoad, line 82: that's a lot of update operations in a loop! in CockroachDB that would absolutely trash the perf, but even PostgreSQL is not going to appreciate it

wonder if a better approach would be to accumulate individual ack/emission facts in redis first, and then do a batch update. and when a job starts, first check if there are any leftovers from the previous attempt

packages/outbox-core/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@message-queue-toolkit/outbox-core",
+  "version": "0.1.0",
```
kibertoad, line 3: 🟢 I'd start at 1.0.0 and follow the semver from there

packages/outbox-core/package.json
```diff
@@ -0,0 +1,63 @@
...
+  },
+  "keywords": [
+    "message",
+    "queue",
+    "queues",
+    "abstract",
+    "common",
+    "utils",
+    "notification",
+    "s3",
```
kibertoad, line 58: s3 doesn't seem relevant

packages/outbox-core/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@message-queue-toolkit/outbox-core",
+  "version": "0.1.0",
+  "private": false,
+  "license": "MIT",
+  "description": "Outbox pattern implementation for message queue toolkit",
```
kibertoad, line 6: can you also add readme?

packages/outbox-core/lib/outbox.ts
kamilwylegala: Sure.

packages/outbox-core/lib/outbox.ts
kamilwylegala: Sure, I didn't notice that. On it.

packages/outbox-core/lib/outbox.ts
kamilwylegala: Sure!

packages/outbox-core/lib/outbox.ts
kamilwylegala: Contract of `getEntries` from `OutboxStorage` interface is supposed to return also all ACKed entires. Such entries are meant to be processed too.

packages/outbox-core/lib/outbox.ts
kamilwylegala: I think that's fine. In my opinion, it's an implementation detail.

E.g. `outbox-prisma-adapter` could accumulate such entries in memory and hit `updateMany` every 20-30 entries.
```js
update(entry) {
  this.entries.push(entry)

  if (this.entries.length > 20) {
    await this.updateMany(this.entries)
    this.entries = []
  }
}
```
The only thing I need to add is to extend `OutboxStorage` with method `flush()` that would be triggered after we finish processing. It could signal to implementation, that there won't be more entries, so whatever is left in `this.entries` could be persisted.

Interface itself is open, also for Redis wrapper for prisma/drizzle adapters. It could be construced in following way:
```js
new OutboxPeriodicJob(
  new RedisOutboxAccumulator(
    diContainer.cradle.redis, //Also we'd need some extra params to configure behavior of accumulator
    new PrismaOutboxAdapter(diContainer.cradle.prisma, 'outbox_entry'), //We provide any impl of OutboxStorage here
  ),  
  eventEmitter,
  dependencies,
  maxRetryCount
  ONE_SEC 
)
```

packages/outbox-core/package.json
kamilwylegala: Could we move to `1.0.0` once API gets stable? 

Regarding the release, I thought we could hold on now with releasing to NPM. Once we have prisma adapter, we could do the release.

And then once it's battle tested in our product, let's promote it to 1.0.0. What do you think?

packages/outbox-core/package.json
kamilwylegala: On it.

packages/outbox-core/package.json
kibertoad: makes sense!

packages/outbox-core/lib/outbox.ts
kibertoad: Then what is the point of ACK step at all, if it's not used for locking?

packages/outbox-core/lib/outbox.ts
kibertoad: > E.g. outbox-prisma-adapter could accumulate such entries in memory and hit updateMany every 20-30 entries.

How would you avoid double publish in case service crashes?

packages/outbox-core/package.json
kamilwylegala: sure, on it

packages/outbox-core/lib/outbox.ts
kibertoad: discussed offline, it is not necessary for now

packages/outbox-core/lib/outbox.ts
kamilwylegala: > How would you avoid double publish in case service crashes?

Also discussed offline, consumers need to take care of it. I will describe everything later in README.md.

kamilwylegala: Done, provided in memory accumulator.

kamilwylegala: Let me also update storage method to have `updateMany`, `updateFailures`.

packages/outbox-core/lib/outbox.ts
```diff
@@ -0,0 +1,127 @@
...
+  public async processOutboxEntries(context: JobExecutionContext) {
+    const entries = await this.outboxStorage.getEntries(this.maxRetryCount)
+
+    await PromisePool.for(entries)
+      .withConcurrency(this.emitBatchSize)
+      .process(async (entry) => {
+        try {
+          await this.eventEmitter.emit(entry.event, entry.data, entry.precedingMessageMetadata)
+          await this.outboxAccumulator.add(entry)
+        } catch (e) {
```
kibertoad, line 37: 🟢 It's a matter of taste, but you can rely on PromisePool for giving you a list of errors:

```ts
const { results, errors } = await PromisePool
  .for([1, 2, 3])
  .withConcurrency(5)
  .useCorrespondingResults()
  .process(async (number) => {
    // …
  })
```

that way try-catch be eliminated. nothing wrong with it, also just a matter of preference

packages/outbox-core/lib/outbox.ts
```diff
@@ -0,0 +1,127 @@
...
+  constructor(
+    private readonly outboxStorage: OutboxStorage<SupportedEvents>,
+    private readonly outboxAccumulator: OutboxAccumulator<SupportedEvents>,
+    private readonly eventEmitter: DomainEventEmitter<SupportedEvents>,
+    private readonly maxRetryCount: number,
+    private readonly emitBatchSize: number,
+  ) {}
+
+  public async processOutboxEntries(context: JobExecutionContext) {
+    const entries = await this.outboxStorage.getEntries(this.maxRetryCount)
```
kibertoad, line 29: can we also check the accumulator in the beginning to see if there is anything already processed there, and filter these out?

packages/outbox-core/lib/storage.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Takes care of persisting and retrieving outbox entries.
+ *
+ * Implementation is required:
+ * - in order to fulfill at least once delivery guarantee, persisting entries should be performed inside isolated transaction
+ * - to return entries in the order they were created (UUID7 is used to create entries in OutboxEventEmitter)
+ * - returned entries should not include the ones with 'SUCCESS' status
+ */
+export interface OutboxStorage<SupportedEvents extends CommonEventDefinition[]> {
+  create(
```
kibertoad, line 14: 1) Maybe method name might be a bit more explicit, so that it's clearer what we are creating? E. g. `createEvent`

2) should it be possible to pass a transaction entity (e. g. Prisma TransactionClient) as a parameter here, so that the transactional guarantee is preserved?

packages/outbox-core/lib/outbox.ts
```diff
@@ -0,0 +1,127 @@
...
+import type { OutboxStorage } from './storage'
+
+/**
+ * Main logic for handling outbox entries.
+ *
+ * If entry is rejected, it is NOT going to be handled during the same execution. Next execution will pick it up.
+ */
+export class OutboxProcessor<SupportedEvents extends CommonEventDefinition[]> {
+  constructor(
+    private readonly outboxStorage: OutboxStorage<SupportedEvents>,
```
kibertoad, line 21: that's a lot of parameters, I would wrap them into two separate config objects -> one for dependencies, one for config

packages/outbox-core/lib/outbox.ts
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * Max retry count is defined by the user.
+ */
+export class OutboxPeriodicJob<
+  SupportedEvents extends CommonEventDefinition[],
+> extends AbstractPeriodicJob {
+  private readonly outboxProcessor: OutboxProcessor<SupportedEvents>
+
+  constructor(
+    outboxStorage: OutboxStorage<SupportedEvents>,
```
kibertoad, line 62: that's also a lot of constructor parameters, wrapping into an object might make it easier to use

packages/outbox-core/lib/outbox.ts
```diff
@@ -0,0 +1,127 @@
...
+      emitBatchSize,
+    )
+  }
+
+  protected async processInternal(context: JobExecutionContext): Promise<void> {
+    await this.outboxProcessor.processOutboxEntries(context)
+  }
+}
+
+export class OutboxEventEmitter<SupportedEvents extends CommonEventDefinition[]> {
```
kibertoad, line 103: sounds a bit like a misnomer, as we are not emitting any events here, we are only scheduling them for future emission

packages/outbox-core/lib/outbox.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    await this.outboxStorage.flush(this.outboxAccumulator)
+    await this.outboxAccumulator.clear()
+  }
+}
+
+/**
+ * Periodic job that processes outbox entries every "intervalInMs". If processing takes longer than defined interval, another subsequent job WILL NOT be started.
+ *
+ * Each entry is ACKed, then event is published, and then entry is marked as SUCCESS. If processing fails, entry is marked as FAILED and will be retried.
```
kibertoad, line 52: is it still acked?

packages/outbox-core/package.json
kamilwylegala: Done, added.

packages/outbox-core/lib/outbox.ts
kamilwylegala: Impemented promise pool.

packages/outbox-core/lib/outbox.ts
kamilwylegala: Updated comment.

packages/outbox-core/lib/storage.ts
kamilwylegala: 1. Sure, `createEvent` sounds good. On it.
2. I think it's up to the client to wrap this with transaction depending on the outbox adapter that will be provided.

For example, following code:
```js
contentLanguageRepository.create(contentLangauge)
eventEmitter.emit('content_langauge.added', payload);
```

It needs to be converted to:
```js
await prisma.$transaction(async () => {
  contentLanguageRepository.create(contentLangauge)
  outboxEmitter.emit('content_langauge.added', payload);
})
```

- Both rows are atomically added to the db.
- Code works with the assumption that `outboxStorage` configured for `outboxEmitter` uses prisma adapter with the same connection like for `contentLanguageRepository`.

Did you have different idea on how to address it?

packages/outbox-core/lib/outbox.ts
kamilwylegala: Sure, I will create separate types for each class:
- Processor
- Job

kamilwylegala: Added `OutboxDependencies` + `OutboxConfiguration`

packages/outbox-core/lib/outbox.ts
kamilwylegala: Addressed too by https://github.com/kibertoad/message-queue-toolkit/pull/204#discussion_r1749829003

packages/outbox-core/lib/outbox.ts
kamilwylegala: Maybe instead let's clear before processing?

kamilwylegala: But it's duplicates concern, right?

Maybe instead we should make sure that accumulator takes care of duplicated values? I.e. each outbox entry has its own unique id, accumulator could check if it already has give entry.

kibertoad: I'm primarily concerned about Redis use-case here. If we use a set for storing elements, we could avoid persisting it to DB twice, but it wouldn't help us to avoid emitting the event twice.

What are the drawbacks of explicit prefiltering?

kamilwylegala: Done.

packages/outbox-core/README.md
```diff
@@ -0,0 +1,42 @@
...
+    //Retry count, how many times outbox entries should be retried to be processed
+    3,
+    //emitBatchSize - how many outbox entries should be emitted at once
+    10,
+    //internalInMs - how often the job should be executed, e.g. below it runs every 1sec
+    1000
+)
+```
+
+Job will take care of processing outbox entries emitted by:
```
kibertoad, line 36: would be nice to write a full usage example for the end user. Now I can see examples of how I can instantiate a bunch of stuff, but have no idea when and how I am supposed to use them.

packages/outbox-core/lib/accumulators.ts
```diff
@@ -0,0 +1,73 @@
...
+  add(outboxEntry: OutboxEntry<SupportedEvents[number]>): Promise<void>
+
+  /**
+   * Accumulates failed event.
+   * @param outboxEntry
+   */
+  addFailure(outboxEntry: OutboxEntry<SupportedEvents[number]>): Promise<void>
+
+  /**
+   * It's meant to be used by OutboxStorage::flush() to get all entries that should be persisted as successful ones.
```
kibertoad, line 25: I would rather document what it does (return all entries accumulated so far) rather than document who and when is expected to use it, as these may vary. E. g. we already use it for prefiltering as well

packages/outbox-core/lib/accumulators.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export class InMemoryOutboxAccumulator<SupportedEvents extends CommonEventDefinition[]>
+  implements OutboxAccumulator<SupportedEvents>
+{
+  private entries: OutboxEntry<SupportedEvents[number]>[] = []
+  private failedEntries: OutboxEntry<SupportedEvents[number]>[] = []
+
+  public add(outboxEntry: OutboxEntry<SupportedEvents[number]>) {
+    this.entries = [...this.entries, outboxEntry]
```
kibertoad, line 49: this is very inefficient. Why not just use `push`?

packages/outbox-core/lib/outbox.spec.ts
```diff
@@ -0,0 +1,327 @@
...
+    originatedFrom: 'service',
+    producedBy: 'producer',
+    schemaVersion: '1',
+    correlationId: randomUUID(),
+  },
+}
+
+const TestLogger: Logger = pino()
+
+class InMemoryOutboxStorage<SupportedEvents extends CommonEventDefinition[]>
```
kibertoad, line 57: feels a bit cluttered with this inlined within the test, can we move it to a separate test directory in its own file?

packages/outbox-core/lib/outbox.ts
```diff
@@ -0,0 +1,139 @@
...
+    private readonly outboxDependencies: OutboxDependencies<SupportedEvents>,
+    private readonly outboxProcessorConfiguration: OutboxProcessorConfiguration,
+  ) {}
+
+  public async processOutboxEntries(context: JobExecutionContext) {
+    const { outboxStorage, eventEmitter, outboxAccumulator } = this.outboxDependencies
+
+    const entries = await outboxStorage.getEntries(this.outboxProcessorConfiguration.maxRetryCount)
+
+    const currentEntriesInAccumulator = new Set(
```
kibertoad, line 45: we can skip this step and the next one altogether if `entries.length === 0`, which is going to be the most typical case

packages/outbox-core/lib/storage.ts
```diff
@@ -0,0 +1,32 @@
...
+   * - Items that are in OutboxAccumulator::getFailedEntries MUST be changed to FAILED status, `updatedAt` field needs to be set and retryCount needs to be incremented.
+   */
+  flush(outboxAccumulator: OutboxAccumulator<SupportedEvents>): Promise<void>
+
+  /**
+   * Returns entries in the order they were created. It doesn't return entries with 'SUCCESS' status. It doesn't return entries that have been retried more than maxRetryCount times.
+   *
+   * For example if entry retryCount is 1 and maxRetryCount is 1, entry MUST be returned. If it fails again then retry count is 2, in that case entry MUST NOT be returned.
+   */
+  getEntries(maxRetryCount: number): Promise<OutboxEntry<SupportedEvents[number]>[]>
```
kibertoad, line 31: would be nice to include a take limit param, but we can make it optional, if you like

packages/outbox-core/README.md
kamilwylegala: Periodic job will start automatically once instantiated, correct?

So I will extend the "emit" part.

packages/outbox-core/lib/accumulators.ts
kamilwylegala: Sure, updated.

packages/outbox-core/lib/accumulators.ts
kamilwylegala: Updated.

packages/outbox-core/lib/outbox.spec.ts
kamilwylegala: Moved test + impl to `test/`

packages/outbox-core/lib/outbox.ts
kamilwylegala: Done,

packages/outbox-core/lib/storage.ts
kamilwylegala: I wanted to make it configurable in prisma adapter. Do you think we will need to change limit across different executions of periodic job?

kibertoad: Sure, that works

kibertoad (approved):
Very impressive work, thank you so much!

kamilwylegala:
> Very impressive work, thank you so much!

Thank you, let me work on the test pipeline now :grin:

kamilwylegala:
Ready for release @kibertoad :tada:

jmontoyaa:
Could we merge this one? 👉👈

jmontoyaa:
Thanks! :pray:  A new version, is not needed at the moment. There's still this PR in draft about the prisma adapter https://github.com/kibertoad/message-queue-toolkit/pull/208  

I will mark ticket AP-5046 as done.

PR #206, opened 2024-09-09
Bugfix/spy for bg listeners

CarlosGamero:


packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -172,46 +162,56 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
-        isSuccessfull = true
-      } finally {
-        this.transactionObservabilityManager?.stop(transactionId, isSuccessfull)
-      }
+      await this.executeEventHandler(event, handler, false)
     }
 
-    for (const handler of eventHandlers.background) {
-      const transactionId = randomUUID()
+    const bgPromises = eventHandlers.background.map((handler) =>
```
kibertoad, line 168: wonder if we should use a promise pool here to have proper concurrency and error handling. we will _probably_ have a small amount of handlers, although that is probably not a guarantee

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -158,9 +147,10 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
    * Register handler for all events supported by the emitter
    */
   public onAny(handler: AnyEventHandler<SupportedEvents>, isBackgroundHandler = false) {
-    for (const supportedEvent of this.eventRegistry.supportedEvents) {
-      this.on(supportedEvent.consumerSchema.shape.type.value, handler, isBackgroundHandler)
-    }
+    const eventTypes = this.eventRegistry.supportedEvents.map(
```
kibertoad, line 150: probably we can calculate this value once during the instantiation? doesn't look like it ever changes

CarlosGamero: Good point! will change it 🙏

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: Not sure if promise pool will help here, we want to start all promises without waiting for them and the error handling is done per promise (if they fail we report the error and continue), maybe I am misunderstanding your message 😓

kibertoad (approved):

PR #207, opened 2024-09-10
AP-5187 Ability to define many sqs prefixes.

kamilwylegala:


packages/sns/lib/sns/AbstractSnsService.ts
```diff
@@ -29,7 +29,7 @@ export type SNSTopicConfig = {
 }
 
 export type ExtraSNSCreationParams = {
-  queueUrlsWithSubscribePermissionsPrefix?: string
+  queueUrlsWithSubscribePermissionsPrefix?: string | string[]
```
kibertoad:
```suggestion
  queueUrlsWithSubscribePermissionsPrefix?: string | readonly string[]
```

kamilwylegala: Done.

kibertoad (approved):

PR #210, opened 2024-09-19
Support a mix of locator and creation configs for SNS

kibertoad:
please review carefully, this was coded in a late night session using the scraps of the mental capacity, likely to be ugly and have gaps.

would be great to clean up the logic and split the code in the follow up prs.

implementation rushed to unblock Billinq service :-/


packages/sns/lib/utils/snsInitter.ts
```diff
@@ -21,9 +30,9 @@ export async function initSnsSqs(
   extraParams?: ExtraParams,
 ) {
   if (!locatorConfig?.subscriptionArn) {
-    if (!creationConfig?.topic) {
+    if (!creationConfig?.topic && !locatorConfig?.topicArn && !locatorConfig?.topicName) {
       throw new Error(
-        'If locatorConfig.subscriptionArn is not specified, creationConfig.topic parameter is mandatory, as there will be an attempt to create the missing topic',
+        'If locatorConfig.subscriptionArn is not specified, creationConfig.topic, localtorConfig.name or locatorConfig.topicArn parameter is mandatory, as there will be an attempt to create the missing topic',
```
kamilwylegala, line 35: Shouldn't we reword the message that attempt to create topic happens only if `creationConfig` is provided?

packages/sns/lib/utils/snsSubscriber.ts
```diff
@@ -19,14 +24,23 @@ export async function subscribeToTopic(
...
+
+  if (!topicArn) {
+    if (isCreateTopicCommand(topicConfiguration)) {
+      topicArn = await assertTopic(snsClient, topicConfiguration, {
+        queueUrlsWithSubscribePermissionsPrefix:
+          extraParams?.queueUrlsWithSubscribePermissionsPrefix,
+        allowedSourceOwner: extraParams?.allowedSourceOwner,
+      })
+    } else {
+      topicArn = await getTopicArnByName(snsClient, topicConfiguration.topicName)
```
kamilwylegala, line 41: If it's not either create command or locate command, shouldn't we throw an error here?

packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
```diff
@@ -85,6 +85,8 @@ export abstract class AbstractSnsSqsConsumer<
         this.creationConfig.queue,
         this.creationConfig.topic,
         this.subscriptionConfig,
+        undefined,
+        this.locatorConfig,
```
kamilwylegala, line 89: TS is complaining now that `initSnsSqs().queueUrl` is string|undefined.

packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
```diff
@@ -12,15 +12,15 @@ import { deleteSnsSqs, initSnsSqs } from '../utils/snsInitter'
...
 
 export type SNSSQSConsumerDependencies = SQSConsumerDependencies & {
   snsClient: SNSClient
 }
 export type SNSSQSCreationConfig = SQSCreationConfig & SNSCreationConfig
 
-export type SNSSQSQueueLocatorType = SQSQueueLocatorType &
-  SNSQueueLocatorType & {
+export type SNSSQSQueueLocatorType = Partial<SQSQueueLocatorType> &
+  SNSTopicLocatorType & {
```
kamilwylegala, line 23: How configuration in billing service could look like?

Arguments passed to constructor should be wrapped with `isTest()` e.g.:
```
{
  "creationConfig": isTest() ? { /* legit creation config */ } : undefined, //In production creation of topics will be disabled.
  "locatorConfig": isTest() ? undefined : { /* legit locate config */ } //In production we will do read-only
}
```

packages/sns/lib/utils/snsInitter.ts
```diff
@@ -68,25 +87,41 @@ export async function initSnsSqs(
...
+    queueName = splitUrl[splitUrl.length - 1]
+  } else {
+    queueName = creationConfig!.queue.QueueName!
+  }
 
   return {
     subscriptionArn: locatorConfig.subscriptionArn,
-    topicArn: locatorConfig.topicArn,
+    topicArn: subscriptionTopicArn,
     queueUrl: locatorConfig.queueUrl,
```
kamilwylegala, line 125: Why does it mix the configuration of both: `locatorConfig` and `creationConfig`? Both can be provided and we return `queueName` from creation config but queue url comes from locate config.

packages/sns/lib/utils/snsInitter.ts
```diff
@@ -97,9 +132,10 @@ export async function deleteSnsSqs(
   snsClient: SNSClient,
   deletionConfig: DeletionConfig,
   queueConfiguration: CreateQueueCommandInput,
-  topicConfiguration: CreateTopicCommandInput,
+  topicConfiguration: CreateTopicCommandInput | undefined,
   subscriptionConfiguration: SNSSubscriptionOptions,
   extraParams?: ExtraParams,
+  topicLocator?: SNSTopicLocatorType,
```
kamilwylegala, line 138: Couldn't we rework this method and introduce a breaking change?

packages/sns/lib/utils/snsInitter.ts
```diff
@@ -150,26 +194,30 @@ export async function deleteSns(
...
-  if (!creationConfig.topic.Name) {
-    throw new Error('topic.Name must be set for automatic deletion')
-  }
+  if (creationConfig.topic) {
+    if (!creationConfig.topic.Name) {
+      throw new Error('topic.Name must be set for automatic deletion')
+    }
 
-  await deleteTopic(snsClient, creationConfig.topic.Name)
+    await deleteTopic(snsClient, creationConfig.topic.Name)
```
kamilwylegala, line 202: Why this change is required?

packages/sns/lib/utils/snsInitter.ts
```diff
@@ -150,26 +194,30 @@ export async function deleteSns(
...
 export async function initSns(
   snsClient: SNSClient,
-  locatorConfig?: SNSQueueLocatorType,
+  locatorConfig?: SNSTopicLocatorType,
   creationConfig?: SNSCreationConfig,
 ) {
   if (locatorConfig) {
-    const checkResult = await getTopicAttributes(snsClient, locatorConfig.topicArn)
+    const topicArn =
+      locatorConfig.topicArn ?? (await getTopicArnByName(snsClient, locatorConfig.topicName))
```
kamilwylegala, line 213: Is it needed? It looks like new API not related to support both creation/locate configs.

kamilwylegala:
EOR1

packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
kibertoad: I'd say that creationConfig would always have an entry for the queue, and conditionally will include entry for the topic.
locatorConfig will conditionally either have an entry for the topic, or be empty

packages/sns/lib/utils/snsInitter.ts
kibertoad: yes

packages/sns/lib/utils/snsInitter.ts
kibertoad: I think we only follow this branch in case we ended up using locate config, but I might be wrong, worth checking

packages/sns/lib/utils/snsInitter.ts
kibertoad: potentially. if you see a clear way to do this that would simplify the code, go for it

packages/sns/lib/utils/snsInitter.ts
kibertoad: which part are you referring to?

packages/sns/lib/utils/snsInitter.ts
kibertoad: it is needed to support both having a predefined arn and a lookup by topic name, which new logic relies upon

packages/sns/lib/utils/snsSubscriber.ts
kibertoad: yeah, this code is complex and hard to follow. would be nice to break into into more reasonable pieces

packages/sns/lib/utils/snsInitter.ts
kamilwylegala: Reworded.

packages/sns/lib/utils/snsSubscriber.ts
kamilwylegala: I read it again more carefully. Actually it's part of the "locate" logic, but instead of having exact topic arn , we use `topicName` to resolve it into `topicArn`. 

Yes, I believe at least "extract method" could be applied here. I'm on it.

kamilwylegala: Done, code adjusted.

packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
kamilwylegala: @kibertoad I solved this by adding `if` to check if queue url is present. From the code it looks it should be mandatory parameter if `subscriptionArn` is provided, we must provide `queueUrl` too. What do you think?

packages/sns/lib/utils/snsInitter.ts
kamilwylegala: OK, so precedence will always be:
- use ARN if present
- then resolve by name (async lookup)

packages/sns/lib/utils/snsInitter.ts
kamilwylegala: There is extra wrapping condition. I'll simplify it.

packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
kamilwylegala: Let me visualize it with the test. I think the rest of the comments will clarify too.

packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
kibertoad: isn't there branching after the check you added, where we skip the queueUrl-related logic if it's not present?

we should support passing queueName instead of queueUrl, no? or we always resolve url based on name later?

kamilwylegala: There is branch with check, but it didn't guarantee that `queueUrl` is present after that:
```ts
let queueName: string
  if ((locatorConfig as SQSQueueLocatorType).queueUrl) {
    const splitUrl = (locatorConfig as SQSQueueLocatorType).queueUrl.split('/')
    queueName = splitUrl[splitUrl.length - 1]
  } else {
    queueName = creationConfig!.queue.QueueName!
  }

  return {
    subscriptionArn: locatorConfig.subscriptionArn,
    topicArn: subscriptionTopicArn,
    queueUrl: locatorConfig.queueUrl,
    queueName,
  }
```
From this code it looks that queueUrl is mandatory.

kibertoad: I don't see it. if queueUrl is not present here, we will just return queueUrl: undefined

kamilwylegala: Yep, but `queueUrl` cannot be null. It's required in `AbstractSnsSqsConsumer`, see line: 104

PR #212, opened 2024-09-30
Feat/bg handler gracefully shutdown

CarlosGamero:


packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -79,6 +82,15 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
     return this._handlerSpy
   }
 
+  public async dispose(): Promise<void> {
+    for (const [key, promise] of this.inProgressBackgroundHandlerByEventId.entries()) {
```
kibertoad, line 86: Why not promise.all?

CarlosGamero: You are absolutely right!

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -162,10 +173,10 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
     }
 
-    const bgPromises = eventHandlers.background.map((handler) =>
-      this.executeEventHandler(event, handler, true),
-    )
-    Promise.all(bgPromises).then(() => {
+    const bgPromise = Promise.all(
+      eventHandlers.background.map((handler) => this.executeEventHandler(event, handler, true)),
+    ).then(() => {
+      this.inProgressBackgroundHandlerByEventId.delete(event.id)
```
kibertoad, line 179: this potentially leaks memory, because promises won't be removed until we call emitter dispose.

Can we add `then` block to every inProgress handler that would remove them from the map, when initiating the process?

kibertoad: wait, no, I misread the code. this is not invoked on disposal, but on processing, so all good.

kibertoad (approved):
great job!

PR #220, opened 2024-10-18
AMQP release prepare for node-core 13

CarlosGamero:
The current released version of AMQP is using node-core 12, and it is producing error on Maestro due to some type discrepancies with InternalError

### DIFF
--- a/packages/amqp/package.json
+++ b/packages/amqp/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@message-queue-toolkit/amqp",
-  "version": "16.1.1",
+  "version": "16.2.0",
   "private": false,
   "license": "MIT",
   "description": "AMQP adapter for message-queue-toolkit",
@@ -25,7 +25,7 @@
     "prepublishOnly": "npm run build:release"
   },
   "dependencies": {
-    "@lokalise/node-core": "^13.0.1",
+    "@lokalise/node-core": "^13.0.0",
     "zod": "^3.23.8"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


dariacm (approved):

packages/amqp/package.json
```diff
@@ -25,7 +25,7 @@
     "prepublishOnly": "npm run build:release"
   },
   "dependencies": {
-    "@lokalise/node-core": "^13.0.1",
+    "@lokalise/node-core": "^13.0.0",
```
kibertoad, line 28: this is not a peer dependency, so I would use the latest available version here, to force the bump

CarlosGamero: Correct me if I am wrong, but this will allow us to be compatible with `13.x.x`, right? using `^13.1.0` it will be `13.1.x`.
I have no strong opinion but I thought that leaving it open to minor updates would be good. If you disagree I will use your suggestion

kibertoad: npm always aim to minimize amount of versions in use. If it can establish a single version that will satisfy constraints from different users, it will use that - e. g. if you have "peerDependency > 12" from package A, "dependency ^13.0.0" from package B and "dependency ^13.1.0" from package C, it will resolve version to 13.1.0, as it satisfies all three constraints.

in this case 13.1.0 is the version where we fixed backwards compatibility (hiddent the symbol that was breaking interface previously), so I think that enforcing it should actually reduce amount of compatibility problems that we will have

CarlosGamero: Thanks for the explanation Igor! it makes a lot of sense, will apply your suggestion and release 🙏

kibertoad (approved):

PR #221, opened 2024-10-18
SNS Tag update

CarlosGamero:


packages/sns/lib/utils/snsUtils.ts
```diff
@@ -178,3 +193,35 @@ export async function getTopicArnByName(snsClient: SNSClient, topicName?: string
...
+  typeof error.Error.Message === 'string' &&
+  error.Error.Code === 'InvalidParameter' &&
+  error.Error.Message.includes('already exists with different tags')
+
+/**
+ * Manually builds the ARN of a topic based on the current AWS account and the topic name.
+ * It follows the following pattern: arn:aws:sns:<region>:<account-id>:<topic-name>
+ * Doc -> https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html
+ */
+const buildTopicArn = async (client: SNSClient, topicName: string) => {
```
kibertoad, line 214: I would expose this from utils, this can be useful in a variety of different use-cases.

Would be nice to also write some tests for it, to ensure that the built arn matches the real arn created by SNS based on same parameters

packages/sns/lib/utils/snsUtils.ts
```diff
@@ -178,3 +193,35 @@ export async function getTopicArnByName(snsClient: SNSClient, topicName?: string
...
+  const region =
+    typeof client.config.region === 'string' ? client.config.region : await client.config.region()
+
+  const stsClient = new STSClient({
```
kibertoad, line 218: recreating client on every response sounds like a lot of overhead, I'd suggest to pass it as a dependency instead

packages/sns/lib/utils/snsUtils.ts
CarlosGamero: Mmm, Okay! will expose it 🙏

packages/sns/lib/utils/snsUtils.ts
CarlosGamero: Do you mean as a dependency for users? the idea was to hide its usage by using SNS config parameters although, I am not really sure 🤔.

What do you mean by recreating it in every response? I think it is done just once on init right?

CarlosGamero: Oh I got it, you mean to the method, would be fine to create it on publisher/consumer creation and pass it as parameter?

kibertoad: wouldn't it be done once per buildTopicArn, which realistically means once for every topic?

kibertoad: yeah, same as we pass sqsClient and snsClient

CarlosGamero: But as a dependency passed by the user, or do you think it is a good idea to build it internally base on SNS config? Advantage of not requesting it is that the API won’t change

CarlosGamero: (Btw sorry, I am already a bit slow 😓)

kibertoad: I would pass by user, would be less surprising than globally cached dependency. but we can also make an optional parameter and build automatically if not passed

PR #222, opened 2024-10-18
SNS Tags update

CarlosGamero:


packages/sns/lib/sns/AbstractSnsService.ts
```diff
@@ -10,6 +11,7 @@ export const SNS_MESSAGE_MAX_SIZE = 256 * 1024 // 256KB
 
 export type SNSDependencies = QueueDependencies & {
   snsClient: SNSClient
+  stsClient: STSClient
```
CarlosGamero, line 14: new dependency as discussed

packages/sns/lib/utils/stsUtils.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  beforeAll(async () => {
+    const diContainer = await registerDependencies({}, false)
+    stsClient = diContainer.cradle.stsClient
+  })
+
+  describe('buildTopicArn', () => {
+    it('build ARN for topic', async () => {
+      const topicName = 'my-test-topic'
+      await expect(buildTopicArn(stsClient, topicName)).resolves.toMatchInlineSnapshot(
```
kibertoad, line 17: might be good to also create a topic with the same name within the test, and compare actual arn to manually built one. then we will break if AWS will change the structure at some point

packages/sns/lib/utils/stsUtils.ts
```diff
@@ -0,0 +1,14 @@
+import { GetCallerIdentityCommand, type STSClient } from '@aws-sdk/client-sts'
+
+/**
+ * Manually builds the ARN of a topic based on the current AWS account and the topic name.
+ * It follows the following pattern: arn:aws:sns:<region>:<account-id>:<topic-name>
+ * Doc -> https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html
+ */
+export const buildTopicArn = async (client: STSClient, topicName: string) => {
+  const identityResponse = await client.send(new GetCallerIdentityCommand({}))
```
kibertoad, line 9: I would move this call outside of the function and make building itself synchronous, but also expose an asynchronous method on top of it that does resolution for you.

from performance perspective it would be ideal to only resolve the identity once per app startup and not once per every queue. for now we can rely on async one, but having sync one exposed as well would allow to optimize in the future by moving identity resolution to app itself

packages/sns/lib/utils/stsUtils.spec.ts
CarlosGamero: Good idea! will add it 🙌 thanks igor 🙇

packages/sns/lib/utils/stsUtils.ts
CarlosGamero: I am not sure if moving identity resolution is a good idea 🤔, maybe we can do it once and cache the result so the next calls won't need it, we can even do it within this file, what do you think?

kibertoad: sounds good! but let's expose a method to reset it as well, for test e. g.

packages/sns/lib/utils/stsUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export const buildTopicArn = async (client: STSClient, topicName: string) => {
+  const identityResponse = await getAndCacheCallerIdentity(client)
+  const region =
+    typeof client.config.region === 'string' ? client.config.region : await client.config.region()
+
+  return `arn:aws:sns:${region}:${identityResponse.Account}:${topicName}`
+}
+
+export const clearCachedCallerIdentity = () => {
```
kibertoad, line 23: let's expose it from package exports as well, because users may need it too

kibertoad (approved):

packages/sns/lib/utils/stsUtils.ts
CarlosGamero: Good point! I forgot to expose it on index 😓

PR #223, opened 2024-10-19
Adding dlq update tags tests

CarlosGamero:


CarlosGamero:
Tests for DLQ tags update were missing, adding them so it is covered

kibertoad (approved):

PR #227, opened 2024-11-05
Ensure background listeners are called even if a foreground listener throws

freiondrej-lokalise:


freiondrej-lokalise:
We might not want to go this way after all - more discussion needed.

PR #231, opened 2024-11-26
Outbox prisma adapter

kamilwylegala:
Transactional outbox pattern adapter for Prisma.


packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
```diff
@@ -22,14 +22,37 @@ export class OutboxPrismaAdapter<SupportedEvents extends CommonEventDefinition[]
...
 
-  flush(outboxAccumulator: OutboxAccumulator<SupportedEvents>): Promise<void> {
-    return Promise.resolve(undefined)
+  async flush(outboxAccumulator: OutboxAccumulator<SupportedEvents>): Promise<void> {
+    const entries = await outboxAccumulator.getEntries()
+
+    const prismaModel: PrismaClient[typeof this.modelName] = this.prisma[this.modelName]
+
+    for (const entry of entries) {
+      await prismaModel.upsert({
```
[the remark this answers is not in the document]
kamilwylegala: Definitely. I'm working on making tests green. Once I cover all cases, Bulk update/insert would do the job.

Actually `upsert` has some weird behavior, maybe let's switch to bulk inserts/updates right away.

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
```diff
@@ -0,0 +1,167 @@
...
+    private readonly prisma: PrismaClient,
+    private readonly modelName: ModelName,
+  ) {}
+
+  createEntry(
+    outboxEntry: OutboxEntry<SupportedEvents[number]>,
+  ): Promise<OutboxEntry<SupportedEvents[number]>> {
+    const prismaModel = this.prisma[this.modelName] as unknown as ModelDelegate
+
+    // @ts-ignore
```
kamilwylegala, line 35: TODO: remove it later, added now to just move forward with the CI and setting up tests there.

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
```diff
@@ -0,0 +1,167 @@
...
+type ModelDelegate = {
+  // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+  create: (args: any) => Promise<any>
+  // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+  findMany: (args: any) => Promise<any>
+  // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+  createMany: (args: any) => Promise<any>
+  // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+  updateMany: (args: any) => Promise<any>
+}
```
CarlosGamero, line 18: 🟡 I am not really happy with using `any` here, as we know the signature of the method could we define entry and output types to have type check and IDE help?
```
type OutboxEntryDelegate<Event extends CommonEventDefinition> = {
  create: (args: { data: OutboxEntry<Event> }) => Promise<OutboxEntry<Event>>
  findMany: (args: { where?: Partial<OutboxEntry<Event>> }) => Promise<OutboxEntry<Event>[]>
....
}
```

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export class OutboxPrismaAdapter<
+  SupportedEvents extends CommonEventDefinition[],
+  ModelName extends keyof PrismaClient & string,
+> implements OutboxStorage<SupportedEvents>
+{
+  constructor(
+    private readonly prisma: PrismaClient,
+    private readonly modelName: ModelName,
+  ) {}
```
CarlosGamero, line 28: 🟡 I remember having issue with passing PrismaClient directly as a parameter on `prima-utils` package, I don't remember exactly the reason, but it was because the object we are using is the autogenerate one and not the default coming from the Prisma package. I fixed it by doing something like:
```
export class OutboxPrismaAdapter<
  SupportedEvents extends CommonEventDefinition[],
  Prisma extends PrismaClient,
  ModelName extends PrismaModelName<Prisma>,
> implements OutboxStorage<SupportedEvents>
{
  constructor(
    private readonly prisma: Prisma,
    private readonly modelName: ModelName,
  ) {}
...
}
```

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  createEntry(
+    outboxEntry: OutboxEntry<SupportedEvents[number]>,
+  ): Promise<OutboxEntry<SupportedEvents[number]>> {
+    const prismaModel = this.prisma[this.modelName] as unknown as ModelDelegate
```
CarlosGamero, line 33: 🟢 I see we are doing this on all methods, could we maybe create a private method to retrieve the model delegate so we can have the casting in a single place

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
kamilwylegala: Thanks for pointing this out. It's still a work in progress. I did `any` here to move quicker. I fully agree, it needs to be addressed. I will work on it.

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
```diff
@@ -0,0 +1,167 @@
...
+          in: [...entries.map((entry) => entry.id), ...failedEntries.map((entry) => entry.id)],
+        },
+      },
+    })
+
+    await this.prisma.$transaction(async (prisma) => {
+      const prismaModel = prisma[this.modelName] as ModelDelegate
+      await this.handleSuccesses(prismaModel, entries, existingEntries)
+      await this.handleFailures(prismaModel, failedEntries, existingEntries)
+    })
```
CarlosGamero, line 67: 🟡 How many entries do we expect here? wondering if chunking might be beneficial

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
```diff
@@ -0,0 +1,167 @@
...
+        },
+      })
+    }
+  }
+
+  private async handleFailures(
+    prismaModel: ModelDelegate,
+    entries: OutboxEntry<SupportedEvents[number]>[],
+    existingEntries: OutboxEntry<SupportedEvents[number]>[],
+  ) {
```
CarlosGamero, line 115: 🟢 This method is almost the same as `handleSuccesses` wondering if we can combine them by adding another param like `isSuccess`

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
```diff
@@ -0,0 +1,167 @@
...
+    private readonly modelName: ModelName,
+  ) {}
+
+  createEntry(
+    outboxEntry: OutboxEntry<SupportedEvents[number]>,
+  ): Promise<OutboxEntry<SupportedEvents[number]>> {
+    const prismaModel = this.prisma[this.modelName] as unknown as ModelDelegate
+
+    // @ts-ignore
+    const messageType = getMessageType(outboxEntry.event)
```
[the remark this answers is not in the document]
CarlosGamero: Of course! let me have a look :)

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
CarlosGamero: Also, I am not sure if the definition of the Model name will work to have IDE autocompletion, wondering if we can make it a but more specific with something like:
```
type PrismaModelName<T extends PrismaClient> = {
  [K in keyof T]: T[K] extends { create: Function; findMany: Function } ? K : never
}[keyof T]

export class OutboxPrismaAdapter<
  SupportedEvents extends CommonEventDefinition[],
  Prisma extends PrismaClient,
  ModelName extends PrismaModelName<Prisma>,
> implements OutboxStorage<SupportedEvents>
{
  constructor(
    private readonly prisma: Prisma,
    private readonly modelName: ModelName,
  ) {}
```

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    const existingEntries = await prismaModel.findMany({
+      where: {
+        id: {
+          in: [...entries.map((entry) => entry.id), ...failedEntries.map((entry) => entry.id)],
+        },
+      },
+    })
+
+    await this.prisma.$transaction(async (prisma) => {
```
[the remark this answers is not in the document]
kamilwylegala: I think I discussed it with Igor, it's kibertoad namespace, so the preference is to not depend on any lokalise pacakges.

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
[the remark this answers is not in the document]
CarlosGamero: Discussed in Slack, For visibility:

Most likely we don't know to save the type in a separate filed and we can save it with the event itself, but just for visibility the issue was that `getMessageType` is expecting a message type and we are passing an event, if I am reading it fine on code, a message is an extension of event and that's why it was failing

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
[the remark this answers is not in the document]
kamilwylegala: Sure, let's use prisma utils.

Back then, it was about using `@lokalise/id-utils`: 

https://github.com/kibertoad/message-queue-toolkit/pull/204#discussion_r1745324442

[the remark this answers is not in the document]
kibertoad: right, it was a slightly different case, we only needed a subset of the library, of which our id-utils is basically an opinionated wrapper

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
kamilwylegala: Fixed

packages/outbox-prisma-adapter/lib/outbox-prisma-adapter.ts
kamilwylegala: Fixed, please take a look :grin:

PR #232, opened 2024-12-11
Concurrent consumers support

kjamrog:
Changes:
- instead of operating on `consumer` field in `AbstractSqsConsumer`, we use `consumers` array now
- number of consumers can be specified with `concurrentConsumersAmount` option
- added `messageProcessingTime` and `messageType` to the log for better observability

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ Multi-schema consumers support multiple message types via handler configs. They 
         * `handlerSpy` - allow awaiting certain messages to be published (see [Handler Spies](#handler-spies) for more information);
         * `logMessages` - add logs for processed messages.
         * `payloadStoreConfig` - configuration for payload offloading. This option enables the external storage of large message payloads to comply with message size limitations of the queue system. For more details on setting this up, see [Payload Offloading](#payload-offloading).
+        * `concurrentConsumersAmount` - configuration for specifying the number of concurrent consumers to create. Available only for SQS consumers
 * `init()`, prepare consumer for use (e. g. establish all necessary connections);
 * `close()`, stop listening for messages and disconnect;
 * `start()`, which invokes `init()`.
--- a/packages/core/lib/queues/AbstractQueueService.ts
+++ b/packages/core/lib/queues/AbstractQueueService.ts
@@ -165,14 +165,22 @@ export abstract class AbstractQueueService<
   }
 
   protected logProcessedMessage(
-    _message: MessagePayloadSchemas | null,
+    message: MessagePayloadSchemas | null,
     processingResult: MessageProcessingResult,
     messageId?: string,
   ) {
+    const messageTimestamp = message ? this.tryToExtractTimestamp(message) : undefined
+    const messageProcessingMilliseconds = messageTimestamp ? Date.now() - messageTimestamp.getTime() : undefined
+
+    // @ts-ignore
+    const messageType = (message && this.messageTypeField in message) ? message[this.messageTypeField] : undefined
+
     this.logger.debug(
       {
         processingResult,
         messageId,
+        messageProcessingTime: messageProcessingMilliseconds,
+        messageType,
       },
       `Finished processing message ${messageId ?? `(unknown id)`}`,
     )
@@ -206,7 +214,6 @@ export abstract class AbstractQueueService<
     if (this.logMessages) {
       // @ts-ignore
       const resolvedMessageId: string | undefined = message?.[this.messageIdField] ?? messageId
-
       this.logProcessedMessage(message, processingResult, resolvedMessageId)
     }
   }
--- a/packages/core/lib/queues/AbstractQueueService.ts
+++ b/packages/core/lib/queues/AbstractQueueService.ts
@@ -170,10 +170,15 @@ export abstract class AbstractQueueService<
     messageId?: string,
   ) {
     const messageTimestamp = message ? this.tryToExtractTimestamp(message) : undefined
-    const messageProcessingMilliseconds = messageTimestamp ? Date.now() - messageTimestamp.getTime() : undefined
+    const messageProcessingMilliseconds = messageTimestamp
+      ? Date.now() - messageTimestamp.getTime()
+      : undefined
 
-    // @ts-ignore
-    const messageType = (message && this.messageTypeField in message) ? message[this.messageTypeField] : undefined
+    const messageType =
+      message && this.messageTypeField in message
+        ? // @ts-ignore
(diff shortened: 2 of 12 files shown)


packages/sqs/test/consumers/SqsPermissionConsumer.spec.ts
```diff
@@ -596,6 +597,69 @@ describe('SqsPermissionConsumer', () => {
...
+        (_, i) => ({
+          id: `${i}`,
+          messageType: 'add',
+          timestamp: new Date().toISOString(),
+        }),
+      )
+
+      await Promise.all(
+        messages.map(async (m) => {
+          await publisher.publish(m)
```
kibertoad, line 651: would be better to do fire-and-forget to reduce the probability of messages just being processed one-by-one by the same consumer

packages/sqs/test/consumers/SqsPermissionConsumer.spec.ts
```diff
@@ -596,6 +597,69 @@ describe('SqsPermissionConsumer', () => {
...
+          await consumer.handlerSpy.waitForMessageWithId(m.id, 'consumed')
```
kibertoad, line 652: it would be better to do this after the messages are already consumed, and not while we are still publishing them

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -165,14 +165,27 @@ export abstract class AbstractQueueService<
   }
 
   protected logProcessedMessage(
-    _message: MessagePayloadSchemas | null,
+    message: MessagePayloadSchemas | null,
     processingResult: MessageProcessingResult,
     messageId?: string,
   ) {
+    const messageTimestamp = message ? this.tryToExtractTimestamp(message) : undefined
```
kibertoad, line 172: I would do the check on the log level for this.logger first and skip it entirely if it's below debug

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -63,6 +63,7 @@ export type SQSConsumerOptions<
     ConsumerOptions,
     'sqs' | 'queueUrl' | 'handler' | 'handleMessageBatch' | 'visibilityTimeout'
   >
+  concurrentConsumersAmount?: number
```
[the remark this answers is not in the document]
kjamrog: Good catch, I forgot about it, already added

packages/sqs/test/consumers/SqsPermissionConsumer.spec.ts
kjamrog: Updated ✅

packages/sqs/test/consumers/SqsPermissionConsumer.spec.ts
kjamrog: Updated ✅

packages/core/lib/queues/AbstractQueueService.ts
kjamrog: From what I see Logger is a generic interface and it does not expose any property for log level. We could check `level` property as it's available in `pino`, but it want apply all of the time

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -96,7 +97,8 @@ export abstract class AbstractSqsConsumer<
   >
   implements QueueConsumer
 {
-  private consumer?: Consumer
+  private consumers: Consumer[]
+  private readonly concurrentConsumersAmount: number
```
CarlosGamero, line 101: 🟢 Do we need to adjust `AbstractSnsSqsConsumer`? I don't know but asking just in case

packages/sqs/test/consumers/SqsPermissionConsumer.spec.ts
```diff
@@ -596,6 +597,66 @@ describe('SqsPermissionConsumer', () => {
     })
   })
 
+  describe('multiple consumers', () => {
+    let diContainer: AwilixContainer<Dependencies>
+    let sqsClient: SQSClient
+
```
CarlosGamero, line 603: 🟢 I would suggest adding a simple test to `AbstractSnsSqsConsumer` just to make sure future changes are not breaking multiple consumer support

CarlosGamero (approved):

packages/core/lib/queues/AbstractQueueService.ts
kibertoad: we should switch to CommonLogger from node-core, it exposes the isLevelEnabled method

README.md
```diff
@@ -100,6 +100,7 @@ Multi-schema consumers support multiple message types via handler configs. They
         * `handlerSpy` - allow awaiting certain messages to be published (see [Handler Spies](#handler-spies) for more information);
         * `logMessages` - add logs for processed messages.
         * `payloadStoreConfig` - configuration for payload offloading. This option enables the external storage of large message payloads to comply with message size limitations of the queue system. For more details on setting this up, see [Payload Offloading](#payload-offloading).
+        * `concurrentConsumersAmount` - configuration for specifying the number of concurrent consumers to create. Available only for SQS consumers
```
kibertoad, line 103:
```suggestion
        * `concurrentConsumersAmount` - configuration for specifying the number of concurrent consumers to create. Available only for SQS and SNS consumers
```

packages/sqs/test/consumers/SqsPermissionConsumer.spec.ts
kjamrog: Similar test added for `AbstractSnsSqsConsumer`

packages/core/lib/queues/AbstractQueueService.ts
kjamrog: So we should probably change it everywhere and since logger is provided from outside that will be breaking change. Should I address it in separate PR then?

kibertoad: yeah, makes sense

packages/sns/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@message-queue-toolkit/sns",
-  "version": "18.0.1",
+  "version": "18.1.1",
```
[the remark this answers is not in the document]
kjamrog: Of course, thanks for catching! Already updated

CarlosGamero (approved):

kibertoad (approved):

PR #233, opened 2024-12-12
Replace Logger with CommonLogger

kjamrog:
Instead of internal `Logger` type, `CommonLogger` interface from `@lokalize/node-core` is used


UPGRADING.md
```diff
@@ -1,5 +1,18 @@
 # Upgrading Guide
 
+## Upgrading from `17.0.0` to `18.0.0`
```
kjamrog, line 3: `core` is updated to `18.0.0` but `sns` and `amqp` are updated to different versions, should we indicate that here?

packages/amqp/package.json
```diff
@@ -25,7 +25,7 @@
     "prepublishOnly": "npm run build:release"
   },
   "dependencies": {
-    "@lokalise/node-core": "^13.1.0",
+    "@lokalise/node-core": "^13.3.0",
```
CarlosGamero, line 28: 🟡 Should we also increase mqt core version from `>=15` to `>=18`?

CarlosGamero: (Same question fro SNS and SQS)

UPGRADING.md
CarlosGamero: I think breaking change is the same for all libraries, right?
Maybe we can do something like:
```
- mqt/core from 17.0.0 to 18.0.0 
- amqp from x to y 
- ...
```

packages/sns/package.json
```diff
@@ -25,7 +25,7 @@
     "prepublishOnly": "npm run build:release"
   },
   "dependencies": {
-    "@lokalise/node-core": "^13.0.1",
+    "@lokalise/node-core": "^13.3.0",
     "sqs-consumer": "^11.1.0",
```
CarlosGamero, line 29: 🟡 I guess we also need to reflect here the new SQS version

CarlosGamero: (On `peerDependencies` I mean, I can add a comment on that line)

packages/amqp/package.json
kjamrog: You mean from compatibility perspective? `CommonLogger` is compatible with old interface so older version should work fine

CarlosGamero: Is it? I am just following the migration guide so please let me know if I missing something:
> If you are using `pino` logger, or any other logger compatible with `CommonLogger` interface, you don't need to do anything. Otherwise, there are serveral properties that need to be provided compared to the old interface:

So I guess it depends on the logger the user is using

kjamrog: That's correct, but it's the case if you want to upgrade to the new version.
So, if you are upgrading, your logger needs to fulfil specific requirements. But if it's already meeting those requirements, it can also be used with old version of the library.

So on this example: if you upgrade `amqp`, but keep old `core`, you will need to use new logger for `amqp` and old `core` will accept it

CarlosGamero: Oh I see, sorry for the confusion 😅 all good, thanks for the explanation 🙏

UPGRADING.md
kjamrog: Yeah, that will be probably more clear, already updated

UPGRADING.md
```diff
@@ -1,5 +1,18 @@
 # Upgrading Guide
 
+## Upgrading </br> `core` `17.0.0` -> `18.0.0` </br> `sqs` `17.0.0` -> `18.0.0` </br> `sns` `18.0.0` -> `19.0.0` </br> `amqp` `16.0.0` -> `17.0.0`    
+
+### Description of Breaking Changes
+- `logger` injectable dependency interface was replaced by `CommonLogger` from `@lokalise/node-core` package, It is compatible with on `pino` logger (https://github.com/pinojs/pino)
```
kibertoad, line 6: is "on" redundant here?

kibertoad (approved):

PR #235, opened 2025-01-02
Message metrics manager dependency

kjamrog:
This PR introduces new `MessageMetricsManager` dependency to `AbstractQueueService`. 

Example usecase: once message is processed, send it's metadata (message type, processing time, result) to `Prometheus`


packages/core/lib/types/queueOptionsTypes.ts
```diff
@@ -5,11 +5,27 @@ import type { PayloadStoreConfig } from '../payload-store/payloadStoreTypes'
...
+
+export type ProcessedMessageMetadata = {
+  messageId: string
+  messageType: string
+  processingResult: MessageProcessingResult
+  messageProcessingMilliseconds?: number
+}
+
+export interface MessageMetricsManager {
+  registerProcessedMessage(metadata: ProcessedMessageMetadata): unknown
```
kibertoad, line 28: do we generally expect this to be synchronous or asynchronous?
why make it `unknown` and not `void`?

kjamrog: Rather synchronous, but even if there would be some async operation inside it's not awaited, so I can declare it as void

kjamrog: Updated ✅

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -216,10 +201,51 @@ export abstract class AbstractQueueService<
...
+    message: MessagePayloadSchemas | null,
+    processingResult: MessageProcessingResult,
+    messageId?: string,
+  ): ProcessedMessageMetadata {
+    // @ts-ignore
+    const resolvedMessageId: string | undefined = message?.[this.messageIdField] ?? messageId
+
+    const messageTimestamp = message ? this.tryToExtractTimestamp(message) : undefined
+    const messageProcessingMilliseconds = messageTimestamp
+      ? Date.now() - messageTimestamp.getTime()
```
kibertoad, line 235: I would set constant to `Date.now` as early as possible, to make the duration calculation more accurate

packages/core/lib/types/queueOptionsTypes.ts
```diff
@@ -5,11 +5,27 @@ import type { PayloadStoreConfig } from '../payload-store/payloadStoreTypes'
...
 
 export type QueueDependencies = {
   errorReporter: ErrorReporter
   logger: Logger
+  messageMetricsManager?: MessageMetricsManager
+}
+
+export type ProcessedMessageMetadata = {
+  messageId: string
+  messageType: string
```
kibertoad, line 22: wonder if we may also want to store a version

packages/amqp/test/utils/testContext.ts
```diff
@@ -208,6 +213,7 @@ export interface Dependencies {
 
   // vendor-specific dependencies
   transactionObservabilityManager: TransactionObservabilityManager
+  messageMetricsManager?: MessageMetricsManager
```
kibertoad, line 216: can you also add documentation about it to readme file?

packages/core/lib/types/queueOptionsTypes.ts
kjamrog: Do you mean `schemaVersion`?

kibertoad: yup!

packages/core/lib/queues/AbstractQueueService.ts
kjamrog: Adjusted

packages/core/lib/types/queueOptionsTypes.ts
kjamrog: `schemaVersion` in metadata is specific to the message format that we use in `DomainEventEmitter`, and we may be interested in other parameters as well, so I added whole message as a property here and made it generic. Let me know if that looks good to you

packages/amqp/test/utils/testContext.ts
kjamrog: added both JSDoc comments and section in README. But I'm not sure if JSDoc comments wouldn't be enough as user only needs to know when the methods are executed and what are the parameters

packages/amqp/test/consumers/AmqpPermissionConsumer.spec.ts
```diff
@@ -111,6 +115,53 @@ describe('AmqpPermissionConsumer', () => {
...
+          messageId: '1',
+          messageType: 'add',
+          processingResult: 'consumed',
+          messageProcessingMilliseconds: expect.any(Number),
+          message: expect.objectContaining({
+            id: '1',
+            messageType: 'add',
+          }),
+        },
+      ])
```
CarlosGamero, line 161: Should we use `consumer.close()` at the end of the test?

packages/amqp/test/consumers/AmqpPermissionConsumer.spec.ts
```diff
@@ -111,6 +115,53 @@ describe('AmqpPermissionConsumer', () => {
     })
   })
 
+  describe('metrics', () => {
```
CarlosGamero, line 118: This test file is quite big already, maybe we should consider splitting it 🤔

CarlosGamero (approved):

packages/amqp/test/consumers/AmqpPermissionConsumer.spec.ts
kjamrog: Good catch, it was missing in few old tests as well, already added

packages/amqp/test/consumers/AmqpPermissionConsumer.spec.ts
kjamrog: For me it's still acceptable size as for test file, but I can split it if you think it will increase readability. In that case I would do the same with `SqsPermissionConsumer.spec.ts` and `SnsSqsPermissionConsumer.spec.ts`, which are even bigger

packages/core/lib/metrics/MessageProcessingTimePrometheusMetric.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Buckets used to configure Histogram metric
+   */
+  buckets: number[]
+
+  /**
+   * Message version used as a label - can be static string or method resolving version based on payload
+   */
+  messageVersion?: string | ((messageMetadata: ProcessedMessageMetadata<T>) => string | undefined)
```
kjamrog, line 27: Message version can be extracted in multiple ways depending on message schema, so allowing to provide it as a parameter

packages/sqs/test/consumers/SqsPermissionConsumer.spec.ts
```diff
@@ -320,6 +320,65 @@ describe('SqsPermissionConsumer', () => {
...
+        ...diContainer.cradle,
+        messageMetricsManager: {
+          registerProcessedMessage(metadata: ProcessedMessageMetadata<SupportedMessages>): void {
+            messagesRegisteredInMetrics.push(metadata)
+          },
+        },
+      })
+      await newConsumer.start()
+
+      publisher.publish({
```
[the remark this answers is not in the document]
kjamrog: Added

README.md
```diff
@@ -371,3 +371,19 @@ For example, in the case of S3, you can set up a lifecycle policy to delete old
   ]
 }
 ```
+
```
kibertoad, line 374: can you also mention the metrics library here?

packages/metrics/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "lint": "biome check . && tsc --project tsconfig.json --noEmit",
+    "lint:fix": "biome check --write .",
+    "test": "vitest",
+    "test:coverage": "npm test -- --coverage",
+    "test:ci": "npm run test:coverage",
+    "prepublishOnly": "npm run build:release"
+  },
+  "dependencies": {
+    "@lokalise/node-core": "^13.3.0",
+    "prom-client": "^15.1.3"
```
kibertoad, line 27: I'd define it as a peerDependency with `>=` range

README.md
kjamrog: Sure, I was planning to fill README after we have final structure and approach. I will also create separate README file for metrics package

packages/metrics/package.json
kjamrog: Good point. We are not using anything specific that would require `15.1.3` version, do you think it would make sense to earlier versions in that case? I wouldn't go below `15.0.0` though, as it was released a year ago already

packages/metrics/lib/prometheus/MessageProcessingTimePrometheusMetric.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (!metadata.messageProcessingMilliseconds) {
+      // Data not available, skipping
+      return
+    }
+
+    this.metric.observe(
+      {
+        messageType: metadata.messageType,
+        version:
+          typeof this.metricParams.messageVersion === 'function'
```
kibertoad, line 63: can we move that conditional resolution to constructor, so that we always invoke a function in this part of the code? (function would either call the one from params, or a dynamically created one just returning the hardcoded version)

packages/metrics/package.json
kibertoad: great idea, let's set `>= 15.0.0`

kjamrog: Updated ✅

README.md
kjamrog: README added ✅

CarlosGamero (approved):

PR #236, opened 2025-01-10
feat(AP-6075): Add publisher & consumer level message deduplication

Drodevbar:
This PR adds publisher-level and consumer-level message deduplication mechanisms.

Changelog:
1. add redis-message-deduplication-store package as ioredis-based deduplication storage (utilizes `redis-sempaphore` under the hood)
2. adjust SQS & SNS publishers and SNS publisher manager to support publisher-level deduplication
3. adjust SQS consumer to support consumer-level deduplication


packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
```diff
@@ -0,0 +1,24 @@
...
+  /** The generator to use for generating deduplication keys for a particular message type */
+  deduplicationKeyGenerator: MessageDeduplicationKeyGenerator
+}
+
+export type MessageDeduplicationConfig = {
+  /** The store to use for storage and retrieval of deduplication keys */
+  deduplicationStore: MessageDeduplicationStore
+
+  /** The configuration for deduplication for each message type */
+  messageTypeToConfigMap: Record<string, MessageDeduplicationMessageTypeConfig>
```
Drodevbar, line 23: It would be great if this, instead of a plain string, accepts message types, so we would have support while defining deduplication config per event type. Unfortunately, I didn't find a way to do so. This is because `AbstractQueueService` allows to provide a custom `messageTypeField` which makes it hard to come out with a proper generic.

 In case you have suggestions how to achieve this, please share.

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export type MessageDeduplicationMessageTypeConfig = {
+  /** How many seconds to keep the deduplication key in the store for a particular message type */
+  deduplicationWindowSeconds: number
+
+  /** The generator to use for generating deduplication keys for a particular message type */
+  deduplicationKeyGenerator: MessageDeduplicationKeyGenerator
+}
+
+export type MessageDeduplicationConfig = {
```
Drodevbar, line 18: The interface might be slightly adjusted once I'll work on consumer-level deduplication. This is because it might be hard for consumers to reuse the same one. For the time being, I've decided not to name it `ProducerMessageDeduplicationConfig`. I'll see how the implementation goes in the case of consumers and adjust/introduce another one.

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
```diff
@@ -0,0 +1,24 @@
+export interface MessageDeduplicationKeyGenerator<Message extends object = object> {
```
eduard-fa, line 1: 🟢🟠 While this approach might work, is it really simpler to require users to create an entirely new class just to generate a key for a message? Wouldn’t it be more straightforward to just accept the key as part of the message metadata? Generating these keys is tightly coupled to knowledge of the message payload. This means, the entity producing the message (i.e., calling eventEmitter.emit()) could just as easily craft the key as part of the message itself.

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -518,4 +524,74 @@ export abstract class AbstractQueueService<
...
     }
   }
+
+  /** Checks for an existence of deduplication key in deduplication store */
+  protected async isMessageDuplicated(message: MessagePayloadSchemas): Promise<boolean> {
+    if (!this.messageDeduplicationConfig) {
+      return false
+    }
+
+    // @ts-ignore
```
eduard-fa, line 534: 🟠 IIRC, `@ts-expect-error` is preferable as the less risky option

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export interface MessageDeduplicationStore {
+  storeKey(key: string, value: string, ttlSeconds: number): Promise<void>
+  retrieveKey(key: string): Promise<string | null>
+}
+
+export type MessageDeduplicationMessageTypeConfig = {
+  /** How many seconds to keep the deduplication key in the store for a particular message type */
+  deduplicationWindowSeconds: number
```
eduard-fa, line 12: 🟢🟠 Just a thought: could this be implemented as a specific case of Redis-based deduplication rather than as a generic solution? 
Consider a scenario where someone wants to use SQS's message deduplication ID instead. In that case, if we would have an abstract enough interfaces, we could provide a straightforward pass-through implementation that maps the deduplication key to the appropriate value based on SQS requirements. For such cases, a deduplication window wouldn’t be necessary, as it wouldn’t be configurable in this context.

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: Adjusted ✅

Drodevbar: Adjusted ✅

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
Drodevbar: What I like about having an extra interface for that is that this approach enforces API users to provide message deduplication keys. 

In case of passing a plain key as metadata while calling the publisher, it might be omitted by mistake (unless I manage to make TS complain about missing dedup key). As an advantage of your proposition, I see a possibility to omit deduplication in certain cases (that could be accomplished with my proposition as well for example by returning null as a key and adjusting queue service to skip dedup in such cases).

I don't have a strong preference. Can rework it to suit your idea. Shall we wait for others' opinions on that?

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
Drodevbar: The case with SQS FIFO and its built-in deduplication mechanism is definitely a valid point. In such a case, we cannot control the deduplication window as AWS doesn't provide such an option. This leads to the conclusion that to make it generic enough it might be wise to move this prop from generic options.

I am wondering where should it live.

Moving it to a specific store level (Redis in our case) would either force us to have the same dedup window for the whole storage or we would need to make the storage aware of event types to have it more granular.

Another option would be to add an extra layer of abstraction which would make it a bit more complex, but would give us more flexibility. 

What do you think about these options?

Drodevbar: Another thought about an extra abstraction layer - maybe instead of making it Redis-specific, we could make it deduplication window-specific. The advantage of that would be that in case of supporting more available storages that also require a deduplication window, API users wouldn't have to provide another integration-deduplication layer but could use a deduplication window-specific one (or the one without it). The con of this approach would be that if we ever run into the need to provide some custom option for a specific storage type, we would anyway end up providing yet another integration-deduplication layer.

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
eduard-fa: Let's hear for other opinions as well, sure

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
eduard-fa: I would probably go with a shared de-duplication service/store instance + a config per producer.
Also, to simplify the use of it, some sort of defaults should be applied. 
I would try to avoid boilerplate code as much as possible.

Drodevbar: After giving it a think again I have a couple of thoughts.

Let's analyze the SQS FIFO example again. It has 2 methods of providing message deduplication.

The first one is content-based deduplication where API consumers don't have to do anything. It just requires setting it up on the queue level and everything is handled automatically. It generates the unique ID on top of the message body.

The second one is by explicitly providing a deduplication ID as one of the message attributes.

Docs: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues-exactly-once-processing.html

That said, the FIFO dedup configuration might vary - in the first scenario nothing is required, in the second one it's just about proving a proper message dedup id.

In case the first option is used, the API consumer doesn't have to provide a deduplication config at all. 

In the case of the second option, MQT already allows passing `SQSMessageOptions` which contains the conditional `MessageDeduplicationId` property. So technically it is possible now to use dedup in case of FIFO queues. 

I think that if we would like to adjust my proposition to support the FIFO case as well, we would have to remove `MessageDeduplicationId` from existing options to unify it and provide only one way of deduplication. In the case of my solution, dedup is heavily store-based but doesn't require modifying the message itself. In the case of FIFO, it is a storage-free solution (since AWS does everything for us), but it requires modifying the message attributes. I am wondering whether we shouldn't think of the proposal as **store-based** deduplication that requires storing the keys somewhere, while FIFO & message dedup ID being something else. 

We can discuss it tomorrow (as agreed on Slack). Just sharing the thoughts from the top of my head so I do not forget.

packages/redis-message-deduplication-store/lib/RedisMessageDeduplicationStore.ts
```diff
@@ -0,0 +1,39 @@
...
+    config: RedisMessageDeduplicationStoreConfig,
+  ) {
+    this.redis = dependencies.redis
+    this.config = config
+  }
+
+  async storeKey(key: string, value: string, ttlSeconds: number): Promise<void> {
+    const keyWithPrefix = this.getKeyWithOptionalPrefix(key)
+
+    await this.redis.set(keyWithPrefix, value, 'EX', ttlSeconds)
```
Drodevbar, line 27: I'll slightly adjust it shortly. I'll use 'NX' to make sure that simultaneous calls are handled properly. Thanks to it `MessageDeduplicationStore` interface will be simplified a bit by requiring only single method implementation. This will also simply AbstractQueueService a bit.

Drodevbar: ✅

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
Drodevbar: I had a huddle with Eduards, during which we discussed an idea for making it more generic and supporting SQS FIFO MessageDeduplicationId-based deduplication as well (despite generic external-store-based deduplication, which was introduced as part of this PR so far).

Here is how it would work:

We could have an abstraction that would allow library users to specify one of the below producer-level message dedup solutions - `StoreBasedMessageDeduplication` and `SqsFifoMessageDeduplicationIdBasedMessageDeduplication`

Depending on what strategy was chosen, the behavior would be different. In the first one it will work as I proposed in this PR so far (so the dedup key will be stored in an external store). In the second one, we'll have to instruct `AbstractSqsPublisher` to adjust the message before sending it in order to inject the `MessageDeduplicationId` attribute. In that approach, we would also have to get rid of the dedup id from `SQSMessageOptions`, because it's currently possible to pass it regardless of whether standard or FIFO is used. 

In case a new deduplication strategy would need to be added, we could just introduce yet another strategy and adjust MQT.
In case of SQS FIFO content-based deduplication, library user will just omit providing the deduplication config as deduplication happens on AWS side.

@kibertoad @Dosexe any thoughts on that? Do you think we need such abstraction now? I was also thinking that to reduce possible confusion, I could just rename the interface now to `StoreBasedMessageDeduplication` so it will be clear to library users that the dedup store-based only.

packages/sns/lib/sns/AbstractSnsPublisher.ts
```diff
@@ -75,6 +75,10 @@ export abstract class AbstractSnsPublisher<MessagePayloadType extends object>
         () => calculateOutgoingMessageSize(updatedMessage),
       )
 
+      if ((await this.deduplicateMessage(parsedMessage)).isDuplicated) {
```
kjamrog, line 78: 🟢 Checking if deduplication is enabled for that message is synchronous, so if we check it first, we can avoid async call. It's minor improvement, but can make some difference in bigger scale.
```suggestion
      if (this.isDeduplicationEnabled(parsedMessage) && (await this.deduplicateMessage(parsedMessage)).isDuplicated) {
```

README.md
```diff
@@ -389,3 +390,67 @@ It needs to implement the following methods:
...
+    import { RedisMessageDeduplicationStore } from '@message-queue-toolkit/redis-message-deduplication-store'
+    import { MessageDeduplicationKeyGenerator } from '@message-queue-toolkit/core'
+
+    const redisClient = new Redis({
+        // your redis configuration
+    })
+
+    // Create a new instance of RedisMessageDeduplicationStore
+    messageDeduplicationStore = new RedisMessageDeduplicationStore(
+      { redis: diContainer.cradle.redis },
```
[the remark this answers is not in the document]
Drodevbar: Good catch, thank you ✅

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
```diff
@@ -0,0 +1,30 @@
...
+   * Stores a deduplication key in case it does not already exist.
+   * Returns true if the key was stored, false if it already existed.
+   */
+  setIfNotExists(key: string, value: string, ttlSeconds: number): Promise<boolean>
+
+  /** Retrieves value associated with deduplication key */
+  getByKey(key: string): Promise<string | null>
+}
+
+export type MessageDeduplicationMessageTypeConfig = {
```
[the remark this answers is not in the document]
Drodevbar: Yes, it could be generic, but I think we won't benefit from passing it down to the key generator. It would be perfect if we could make keys generic, but I think it's not possible due to https://github.com/kibertoad/message-queue-toolkit/pull/236#discussion_r1914553442

Let me anyway do it like you suggested to make it more consistent ✅

packages/sns/lib/sns/AbstractSnsPublisher.ts
Drodevbar: Great point, thanks, addjusted ✅

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
Drodevbar: > or having single value per publisher (could be a function to give more flexibility)

In my opinion, having this option on the publisher level would decrease the usability of it. I can imagine that various events that publisher support might require different dedup windows.

Since you're also leaning towards Ed's approach, I think we could have both at message metadata level - key (or simple key generator function) and the dedup window.

Since there are 2 votes for this approach, let me rework it shortly

README.md
```diff
@@ -389,3 +390,67 @@ It needs to implement the following methods:
...
+    import { MessageDeduplicationKeyGenerator } from '@message-queue-toolkit/core'
+
+    const redisClient = new Redis({
+        // your redis configuration
+    })
+
+    // Create a new instance of RedisMessageDeduplicationStore
+    messageDeduplicationStore = new RedisMessageDeduplicationStore(
+      { redis: redisclient },
+      { keyPrefix: 'optional-key-prefix' }, // used to prefix deduplication keys
```
kibertoad, line 423: doesn't redis client by itself support keyprefix?

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -518,4 +524,67 @@ export abstract class AbstractQueueService<
...
+    message: MessagePayloadSchemas,
+  ): Promise<{ isDuplicated: boolean }> {
+    if (!this.messageDeduplicationConfig) {
+      return { isDuplicated: false }
+    }
+
+    // @ts-expect-error
+    const messageType = message[this.messageTypeField] as string
+
+    if (!this.messageDeduplicationConfig.messageTypeToConfigMap[messageType]) {
```
kibertoad, line 553: why not use `isDeduplicationEnabled` here? entire block seems duplicated

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -518,4 +524,67 @@ export abstract class AbstractQueueService<
...
+      return { isDuplicated: false }
+    }
+
+    const deduplicationKeyStored =
+      await this.messageDeduplicationConfig.deduplicationStore.setIfNotExists(
+        this.messageDeduplicationConfig.messageTypeToConfigMap[
+          messageType
+        ].deduplicationKeyGenerator.generate(message),
+        new Date().toISOString(),
+        this.messageDeduplicationConfig.messageTypeToConfigMap[messageType]
```
kibertoad, line 563: we use this variable twice, I would store it to a const value prior to calling `setIfNotExists` to simplify the code

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -518,4 +524,67 @@ export abstract class AbstractQueueService<
...
+    }
+
+    // @ts-expect-error
+    const messageType = message[this.messageTypeField] as string
+
+    if (!this.messageDeduplicationConfig.messageTypeToConfigMap[messageType]) {
+      return { isDuplicated: false }
+    }
+
+    const deduplicationKeyStored =
```
kibertoad, line 557: if this is a boolean, I would call it `wasDeduplicationKeyStored` for clarity

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -518,4 +524,67 @@ export abstract class AbstractQueueService<
...
+        ].deduplicationKeyGenerator.generate(message),
+        new Date().toISOString(),
+        this.messageDeduplicationConfig.messageTypeToConfigMap[messageType]
+          .deduplicationWindowSeconds,
+      )
+
+    return { isDuplicated: !deduplicationKeyStored }
+  }
+
+  private getValidateMessageDeduplicationConfig(
```
kibertoad, line 570: `getValidatedXYZ`, perhaps?

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -518,4 +524,67 @@ export abstract class AbstractQueueService<
...
+      return undefined
+    }
+
+    for (const messageConfig of Object.values(messageDeduplicationConfig.messageTypeToConfigMap)) {
+      const messageTypeToConfigMapParseResult =
+        MESSAGE_DEDUPLICATION_MESSAGE_TYPE_SCHEMA.safeParse(messageConfig)
+
+      if (messageTypeToConfigMapParseResult.error) {
+        throw new Error(
+          `Invalid message deduplication config provided: ${messageTypeToConfigMapParseResult.error.toString()}`,
```
kibertoad, line 583: what is the type of error here? will it toString correctly? (objects don't)

README.md
kibertoad: or is the idea here that we may want to have different prefixes on the same connection?

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
kibertoad: I don't have a strong opinion on this. It is unlikely that we will use FIFO any time soon, but if it's relative simple to have this extension point for the future, I don't mind it.

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
kibertoad: >  I can imagine that various events that publisher support might require different dedup windows.

I agree with this concern. As we have one publisher per topic, we can't really predict how difficult event types within that topic will be.

packages/redis-message-deduplication-store/lib/RedisMessageDeduplicationStore.ts
```diff
@@ -0,0 +1,40 @@
...
+    dependencies: RedisMessageDeduplicationStoreDependencies,
+    config: RedisMessageDeduplicationStoreConfig,
+  ) {
+    this.redis = dependencies.redis
+    this.config = config
+  }
+
+  async setIfNotExists(key: string, value: string, ttlSeconds: number): Promise<boolean> {
+    const keyWithPrefix = this.getKeyWithOptionalPrefix(key)
+    const result = await this.redis.set(keyWithPrefix, value, 'EX', ttlSeconds, 'NX')
```
kibertoad, line 26: what if we error out here? Wonder if we may want to try-catch and resolve to false in that case, or fail fast and break the flow

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
Drodevbar: @kibertoad Currently the configuration is defined on a publisher level (or publisher manager but then it's passed down to publishers as well). Do you think it's worth allowing to specify it at both publisher and event levels? What would that solve for us? Krzysztof and Eduards were leaning towards moving the config from publisher level to event level. I was also thinking about this idea.

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
Drodevbar: Okay, if you don't mind, I'll then mark the current deduplication as store-based and explicitly mark in docs that for FIFO appropriate message attribute should be used (or queue should be configured in a proper way) instead of store-based one.

README.md
Drodevbar: I was thinking that we could use the same Redis connection as part of a service for both "regular" operations (like tmp storage management for different types of needs) and deduplication. In such a case, I think we need to handle key prefixing on dthe edup store. Feel free to disagree.

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: It's because I need to retrieve `messageType` (or being more specific - a configuration for a specific message type), while `isDeduplicationEnabled` returns a boolean. I'll introduce another function that will be reduced as part of this one and `isDeduplicationEnabled` so we can share the logic of retrieving it.

Drodevbar: Adjusted the code ✅

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: Adjusted ✅

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: Adjusted ✅

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: Absolutely, it's a typo, already corrected as part of consumer PR which I merged into this one a while ago. ✅

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: If you're referring to `messageTypeToConfigMapParseResult.error`, it's a zod one coming from a schema validation error. The whole error outputs something as below after calling `toString()`

```txt
Error: Invalid consumer message deduplication config provided: [
  {
    "code": "too_small",
    "minimum": 0,
    "type": "number",
    "inclusive": false,
    "exact": false,
    "message": "Number must be greater than 0",
    "path": [
      "deduplicationWindowSeconds"
    ]
  },
  {
    "code": "too_small",
    "minimum": 0,
    "type": "number",
    "inclusive": false,
    "exact": false,
    "message": "Number must be greater than 0",
    "path": [
      "maximumProcessingTimeSeconds"
    ]
  }
]
```

packages/redis-message-deduplication-store/lib/RedisMessageDeduplicationStore.ts
Drodevbar: If we want to handle this specific case, should I handle all of cases when we're communicating with redis to gracefully handle connection errors?

README.md
kibertoad: do we need a user-defined key for this? wonder if there could be a default prefix per use-case, with a possibility for user to override

README.md
```diff
@@ -389,3 +390,67 @@ It needs to implement the following methods:
   - `messageProcessingMilliseconds` - message processing time in milliseconds
 
 See [@message-queue-toolkit/metrics](packages/metrics/README.md) for concrete implementations
+
+## Publisher-level message deduplication
+
+Publisher-level message deduplication is a mechanism that prevents the same message from being sent to the queue multiple times.
+It is useful when you want to ensure that a message is published only once, regardless of how many times it is sent.
+
+Note that in case of some queuing systems, such as standard SQS, publisher-level deduplication is not sufficient to guarantee that a message is **processed** only once.
```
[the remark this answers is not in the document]
Drodevbar: Adjusted ✅

packages/amqp/lib/AbstractAmqpConsumer.ts
```diff
@@ -268,6 +268,10 @@ export abstract class AbstractAmqpConsumer<
     )
   }
 
+  protected override queueMessageForRetry(message: MessagePayloadType): void {
+    this.channel.nack(message as Message, false, true)
+  }
```
CarlosGamero, line 273: 🟢 Should this be private?

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
```diff
@@ -0,0 +1,74 @@
...
+  /** The generator to use for generating deduplication keys for a particular message type */
+  deduplicationKeyGenerator: MessageDeduplicationKeyGenerator<Message>
+}
+
+export type ConsumerMessageDeduplicationConfig = {
+  /** The store to use for storage and retrieval of deduplication keys */
+  deduplicationStore: ConsumerMessageDeduplicationStore
+
+  /** The configuration for deduplication for each message type */
+  messageTypeToConfigMap: Record<string, ConsumerMessageDeduplicationMessageTypeConfig>
```
CarlosGamero, line 79: 🟢 This is pretty much the same as `PublisherMessageDeduplicationConfig`, maybe we can simplify it with
```
 type MessageDeduplicationConfig<
  T extends ConsumerMessageDeduplicationStore | PublisherMessageDeduplicationStore,
  U extends ConsumerMessageDeduplicationMessageTypeConfig | PublisherMessageDeduplicationConfig,
> = {
  deduplicationStore: T
  messageTypeToConfigMap: Record<string, U>
}
```

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
```diff
@@ -0,0 +1,74 @@
...
+export type ConsumerMessageDeduplicationMessageTypeConfig<Message extends object = object> = {
+  /** How many seconds to keep the deduplication key in the store for a particular message type after message is successfully processed */
+  deduplicationWindowSeconds: number
+
+  /** How many seconds it is expected to take to process a message of a particular type */
+  maximumProcessingTimeSeconds: number
+
+  /** The generator to use for generating deduplication keys for a particular message type */
+  deduplicationKeyGenerator: MessageDeduplicationKeyGenerator<Message>
+}
```
CarlosGamero, line 72: 🟢 
```
type ConsumerMessageDeduplicationMessageTypeConfig<Message extends object = object> =
  PublisherMessageDeduplicationMessageTypeConfig<Message> & {
    maximumProcessingTimeSeconds: number
  }
```

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -518,4 +536,158 @@ export abstract class AbstractQueueService<
...
+
+    // Message was already processed
+    if (deduplicationKeyStatus === ConsumerMessageDeduplicationKeyStatus.PROCESSED) {
+      return false
+    }
+
+    // Message is still being processed within the expected time
+    // Queue it for next check
+    await this.queueMessageForRetry(message)
+    return false
```
CarlosGamero, line 611: 🟢 Most likely this is a very stupid question, and the code is right, but just for my understanding 😓. If the message is already being processed, does it mean that we have the same message twice? if so, why do we need to retry the duplicate?

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -518,4 +536,158 @@ export abstract class AbstractQueueService<
...
+  ): ConsumerMessageDeduplicationMessageTypeConfig | undefined {
+    if (!this.consumerMessageDeduplicationConfig) {
+      return undefined
+    }
+
+    // @ts-expect-error
+    const messageType = message[this.messageTypeField] as string
+
+    return this.consumerMessageDeduplicationConfig.messageTypeToConfigMap[messageType] ?? undefined
+  }
```
CarlosGamero, line 669: 🟢 Could we merge both methods? Similar to what we are doing on method below
```
private getDeduplicationConfigForMessage<T extends ConsumerMessageDeduplicationConfig | PublisherMessageDeduplicationConfig>(
    message: MessagePayloadSchemas,
    deduplicationMessageTypeConfig: T | undefined,
  ): T['messageTypeToConfigMap'][string] | undefined {
    ...
  }
```

packages/sns/lib/sns/SnsPublisherManager.messageDeduplication.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+        creationConfig: {
+          updateAttributesIfExists: true,
+        },
+        producerMessageDeduplicationConfig: {
+          deduplicationStore: messageDeduplicationStore,
+          messageTypeToConfigMap: {
+            'entity.created': {
+              deduplicationKeyGenerator: messageDeduplicationKeyGenerator,
+              deduplicationWindowSeconds: 10,
+            },
```
CarlosGamero, line 65: Just for my knowledge.

We need to provide a config per event type, how are we going to handle this on the Publisher manager, where all publishers are created using a Publisher factory?

packages/sns/lib/sns/SnsPublisherManager.messageDeduplication.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+            // 'entity.update' is not configured on purpose
+          },
```
CarlosGamero, line 67: Do you know how we are going to integrate this with Flow? I mean, we need to define the config per message type and I am wondering if to have a default config, we would need some changes on AbstractPublisherManager? I don't know, asking just in case

CarlosGamero:
Awesome job @Drodevbar 🚀! And sorry for joining so late to the review party 😓

packages/core/lib/message-deduplication/messageDeduplicationSchemas.ts
```diff
@@ -0,0 +1,6 @@
+import { z } from 'zod'
+
+export const MESSAGE_DEDUPLICATION_MESSAGE_TYPE_SCHEMA = z.object({
+  deduplicationWindowSeconds: z.number().int().gt(0).optional(),
+  maximumProcessingTimeSeconds: z.number().int().gt(0).optional(),
+})
```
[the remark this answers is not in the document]
Drodevbar: You're absolutely right - the schema is fully in line with consumer config and to some point with publisher one (`deduplicationWindowSeconds` is present thus I made the fields optional)

This schema is basically used to validate the config upon consumer/publisher creation.

One thing that bothers me with switching to your approach is loosing jsdoc support. I am no sure how important it us fo us, but I personally do like to have it in place to get some hints on the params I am referring to. Maybe there is a way to keep jsdoc in place while using zod as well, but I haven't figured it out (i tried with describe or annotating the field in the schema).

Hope the pictures make it clear what i mean by the above.

current behavior
Screenshot 2025-01-22 at 21 40 39

with creating types on top of zod schemas
Screenshot 2025-01-22 at 21 36 48
Screenshot 2025-01-22 at 21 36 31

> Asking because if we change the type but not the schema, it might cause issues with the parsing operation

Btw, there is a test that checks consumer/publisher init with wrong config provided - it should prevent this case.

I don't have a strong preference on that, I can adjust if you think it's worth reusing the types from zod 🙌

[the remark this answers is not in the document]
CarlosGamero: That's a nice point! I didn't think about the JSDoc and I believe it is beneficial.

I have been playing a bit with Zod and I found a way to preserve JSDoc, although I am not very happy with it, it is the only way that worked for me.

Screenshot 2025-01-23 at 09 38 27

What do you think?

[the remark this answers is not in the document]
Drodevbar: If you found a way to make it work, sure - let's do it. Thanks for checking it our!

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: Absolutely valid question! 😊  
> If the message is already being processed, does it mean that we have the same message twice?  

I can imagine we could encounter this situation in two scenarios:  
1. **The message is a duplicate**: It’s already being processed by another consumer (as you mentioned).  
2. **The message is not a duplicate**: The consumer experienced a fatal failure while processing the message (e.g., the app was restarted or crashed). Upon restarting, the deduplication key with the maximum processing time still exists. In this case, we don’t want to lose the message, but instead, we’d want to wait for the deduplication key to expire. This is the only scenario this approach would address.  

Feel free to challenge this perspective. Point 2 seems like a niche edge case that might not be worth addressing. Point 1, on the other hand, doesn’t require re-queuing the message.

kibertoad: I would challenge the assumption that scenario 2 is not worth addressing. apps crash randomly without warning all the time, assuming app exists long enough it's almost guaranteed to happen. we absolutely must handle this case.

wouldn't original idea to store processing status within the lock address it?

packages/sns/lib/sns/SnsPublisherManager.messageDeduplication.spec.ts
Drodevbar: It's possible to pass the config through the publisher manager. You can see the tests of it to see how it looks. The config is per message type, that's correct. I expect this will slightly change as per this discussion https://github.com/kibertoad/message-queue-toolkit/pull/236#discussion_r1914598783 we probably want to get rid of the key generator and allow passing it explicitly when publishing a message. It might be the same story with dedup window. I'll see how it goes once I refactor it 🙌 Let me know if you have any more questions on that topic.

packages/core/lib/queues/AbstractQueueService.ts
CarlosGamero: I see! and I agree that makes sense to contemplate the second scenario to be safe. My only suggestion here, then, is to add a small comment there explaining this to avoid future questions/misunderstandings

packages/sns/lib/sns/SnsPublisherManager.messageDeduplication.spec.ts
CarlosGamero: Makes sense! thanks for the info 🙏

README.md
Drodevbar: Keep in mind that prefix is optional, I can add a default one in case user doesn't specify it (but in such case maybe they don't need prefixes at all? 🤔)

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: @CarlosGamero Sure, will add a comment

@kibertoad 
> wouldn't original idea to store processing status within the lock address it?

There are statuses in place as part of this solution - `PROCESSING` and `PROCESSED`. I think these 2 + 2 different TTLs (one expected processing time vs dedup window) should be sufficient to handle the possible scenarios.

This is the current flow:
Screenshot 2025-01-23 at 17 09 23

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
Drodevbar: This one I would leave this as is if you don't mind. `maximumProcessingTimeSeconds` is presented in both types, but it has a slightly different meaning (on consumer it is a TTL that's applied once the message is successfully processed, while on publisher level it's applied before publishing a message). I would like to have a separate JSDoc describing the difference (If I extend one from another, JSDoc will be the same for both afaik). Btw, this will be anyway sitting in zod schema (as per you previous suggestion 🙌 )

packages/core/lib/message-deduplication/messageDeduplicationSchemas.ts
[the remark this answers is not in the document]
Drodevbar: Adjusted ✅

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export type PublisherMessageDeduplicationConfig = {
+  /** The store to use for storage and retrieval of deduplication keys */
+  deduplicationStore: PublisherMessageDeduplicationStore
+
+  /** The configuration for deduplication for each message type */
+  messageTypeToConfigMap: Record<string, PublisherMessageDeduplicationMessageTypeConfig>
+}
+
+export interface ConsumerMessageDeduplicationStore {
```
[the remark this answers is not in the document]
Drodevbar: Adjusted ✅

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
Drodevbar: Adjusted ✅

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: Adjusted ✅

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
Drodevbar: @eduard-fa @kjamrog Hey, I've removed the idea of key generator and replaced it with passing `deduplicationId` as message metadata ✅

packages/redis-message-deduplication-store/lib/RedisMessageDeduplicationStore.ts
Drodevbar: @kibertoad I've added handling of dedup-related errors on `AbstractQueueService` level. In case of an error, it is handled by `handleError` (method that was already in place, it logs the errors and reports it) and the message is being processed with deduplication turned off. Let me know if that's a good approach to you.

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -518,4 +561,184 @@ export abstract class AbstractQueueService<
...
+
+  protected isConsumerDeduplicationEnabled(message: MessagePayloadSchemas): boolean {
+    return !!this.getDeduplicationConfigForMessage(message, this.consumerMessageDeduplicationConfig)
+  }
+
+  /**
+   * Tries to acquire lock for processing the message.
+   * Returns true in case if lock has been acquired and message can be processed. Returns false otherwise.
+   */
+  protected async tryToAcquireLockForProcessing(message: MessagePayloadSchemas): Promise<boolean> {
```
Drodevbar, line 616: @kibertoad I'm copy-posting your suggestion from the previous PR https://github.com/kibertoad/message-queue-toolkit/pull/237#discussion_r1920105956

> can we move the lock management logic inside the deduplication store then? (primarily move setIfNotExists and getByKey methods inside, and expose tryToAcquireLockForProcessing method from the store interface itself)

I am wondering if this is the right approach, but please feel free to correct me. `tryToAcquireLockForProcessing` (and other dedup-related methods defined on `AbstractQueueService` level) have some logic which I initially wanted to keep the same regardless of what deduplication store is used. I thought that if someone would like to use some custom dedup store, they just would need to provide the basic methods defined by `MessageDeduplicationConfig`, but the user wouldn't have to provide custom dedup logic. If we move `tryToAcquireLockForProcessing` to the dedup store level, there is a risk that someone will provide an invalid implementation which might negatively impact deduplication logic. Also, there are a couple of methods used as part of this method that are available only to `AbstractQueueService` - this could be of course tackled by allowing to pass custom callbacks, but still I am not sure if this is the right direction. Feel free to disagree if you think it's worth moving it (and possible other dedup-related methods) to dedup store.

README.md
```diff
@@ -389,3 +391,135 @@ It needs to implement the following methods:
...
 
 See [@message-queue-toolkit/metrics](packages/metrics/README.md) for concrete implementations
+
+## Publisher-level store-based message deduplication
+
+Publisher-level store-based message deduplication is a mechanism that prevents the same message from being sent to the queue multiple times.
+It is useful when you want to ensure that a message is published only once, regardless of how many times it is sent.
+
+The mechanism relies on a deduplication store, which is used to store deduplication keys for a certain period of time.
+Before message is published, a deduplication key is generated based on the message content and checked against the store.
```
kibertoad, line 401: I would suggest to disambiguate this a bit:
1) Mention that deduplication key must be generated and passed ("is generated" kinda implies that MQT will generate it);
2) Explain what makes a good deduplication key.

README.md
kibertoad: it's OK to have prefix, but then let's explain what it's supposed to be and when different values need to be passed, so that user is not confused

README.md
```diff
@@ -389,3 +391,135 @@ It needs to implement the following methods:
...
+    }
+    ```
+
+## Consumer-level store-based message deduplication
+
+Consumer-level store-based message deduplication is a mechanism that prevents the same message from being processed multiple times.
+It is useful when you want to be sure that message is processed only once, regardless of how many times it is received.
+
+The mechanism relies on a deduplication store, which is used to store deduplication keys for a certain period of time.
+Upon processing a message, a deduplication key is generated based on the message content and checked against the store.
```
kibertoad, line 460: I assume in this case consumer itself generates a deduplication key? or it relies on the one that publisher has generated? let's make that explicit

README.md
```diff
@@ -389,3 +391,135 @@ It needs to implement the following methods:
...
+In case the key doesn't exist in the store, key is stored with status `PROCESSING` in the store for a certain period of time (i.e. `maximumProcessingTimeSeconds`) and the message is processed.
```
kibertoad, line 461: let's document what will happen if another consumer will try to process a message with the same deduplicationId while it already exists in store with status `PROCESSING`

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -518,4 +561,184 @@ export abstract class AbstractQueueService<
...
+
+    if (!messageDeduplicationConfig || !deduplicationId) {
+      return true
+    }
+
+    const consumerDeduplicationConfig = this
+      .consumerMessageDeduplicationConfig as ConsumerMessageDeduplicationConfig
+
+    try {
+      const result = await consumerDeduplicationConfig.deduplicationStore.setIfNotExists(
```
kibertoad, line 632: I would make the name more informative, e. g. `wasLockAcquired`

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -518,4 +561,184 @@ export abstract class AbstractQueueService<
...
+        return false
+      }
+
+      // Message is still being processed within the expected time
+      // Queue it for next check
+      await this.queueMessageForRetry(message)
+      return false
+    } catch (err) {
+      this.handleError(err)
+      return true
```
kibertoad, line 657: is this intentional, that in case of error we return `true`? let's add comment explaining what will happen

packages/core/lib/queues/AbstractQueueService.ts
kibertoad: Main benefit of using redis-semaphor is built-in handling of long execution with the auto-extension of lock if service is still alive and kicking, but just taking more time than originally expected. It is not a complex logic, however, we can implement it on our end as well, in store-agnostic fashion, most likely.

Check https://github.com/swarthy/redis-semaphore/blob/master/src/Lock.ts out for the `startRefresh`/`stopRefresh` functions.

packages/redis-message-deduplication-store/lib/RedisMessageDeduplicationStore.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  constructor(
+    dependencies: RedisMessageDeduplicationStoreDependencies,
+    config: RedisMessageDeduplicationStoreConfig,
+  ) {
+    this.redis = dependencies.redis
+    this.config = config
+  }
+
+  async setIfNotExists(key: string, value: string, ttlSeconds: number): Promise<boolean> {
```
kibertoad, line 29: wonder if we can reduce the amount of hops by adding `GET` flag to the `set` operation, which will also return the value if it already exists.

might not be worth it, because lock already existing is unlikely to be a hot path.

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
```diff
@@ -0,0 +1,57 @@
...
+    | PublisherMessageDeduplicationMessageTypeConfig,
+> = {
+  /** The store to use for storage and retrieval of deduplication keys */
+  deduplicationStore: TStore
+
+  /** The configuration for deduplication for each message type */
+  messageTypeToConfigMap: Record<string, TConfig>
+}
+
+export enum ConsumerMessageDeduplicationKeyStatus {
```
kibertoad, line 54: what if message fails with "Error, do not retry"? (e. g. payload cannot be parsed or does not pass validation).

Are we / should we mark the message as `PROCESSED`?

README.md
Drodevbar: Good point, I forgot to update it after switching from key generator approach to passing the dedup id directly. Will change

README.md
Drodevbar: Added the description to `RedisMessageDeduplicationStoreConfig`

```ts
export type RedisMessageDeduplicationStoreConfig = {
  /** Optional prefix for the keys in the store, useful for namespacing and differentiating between deduplication store and other data */
  keyPrefix?: string
}
```

README.md
Drodevbar: It's the latter one - the dedup id is included in the message metadata. I've updated the docs:

```txt
The mechanism relies on:
1. a deduplication store, which is used to store deduplication keys for a certain period of time.
2. a deduplication key, which should be stored as part of the message before publishing it (`messageDeduplicationIdField` property of `newPublisherOptions` lets you specify the field in the message that contains the deduplication id)

In case the key doesn't exist in the store, key is stored with status `PROCESSING` in the store for a certain period of time (i.e. `maximumProcessingTimeSeconds`) and the message is processed.
Upon successful processing, deduplication key TTL is updated to `deduplicationWindowSeconds` to prevent processing the same message again. The value is updated to `PROCESSED`.
In case of errors handled gracefully by a consumer, deduplication key is removed from the store to allow instant reprocessing of the message.
In case of unexpected errors, the message will be retried again after deduplication key TTL expires.
In case of another consumer trying to process the same message while its status is `PROCESSING`, the message will be retried again according to `queueMessageForRetry` method which can be override by the consumer.
In case of another consumer trying to process the same message while its status is `PROCESSED`, the message will be considered as a duplicate and will be ignored.
```

README.md
Drodevbar: Updated the docs as per https://github.com/kibertoad/message-queue-toolkit/pull/236#discussion_r1932340975

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: Absolutely, thank you, adjusted

packages/redis-message-deduplication-store/lib/RedisMessageDeduplicationStore.ts
Drodevbar: AFAIK `set` with `GET` was introduced as part of redis 7. I see in AP we're using redis version `6.x`. I could use `multi` as part of `setIfNotExists` (or rather `setIfNotExistsAndReturnValue`), but this will probably not be the hot path as you mentioned.

packages/core/lib/message-deduplication/messageDeduplicationTypes.ts
Drodevbar: I see below edge cases that I haven't thought about before:
1. message validation fails and we remove the message from the queue
2. `retryLater` and message should be retried
3. `retryLater` and message should not be retried (due to exceeding the retries)
4. error during processing the message 

Currently, deduplication only takes into consideration point 4. I think we should react to some of the previous cases as well.

How about the below?
1. mark the message as `PROCESSED`
2. remove the deduplication key to avoid retrying upon the next consumer invocation
3. mark the message as `PROCESSED`

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: Absolutely, thank you, adjusted

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: My idea was to allow the message to be processed in case of dedup store-related errors. Since this method resolves to `true`/`false` depending on if the lock was acquired, `true` here means that we allow further processing. I've added the comment.

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: @kibertoad I'll adjust, thank you for the clarification 🙇

I am also wondering if it makes sense to enqueue the message in case we see another consumer is processing it. Maybe with the usage of `redis-semaphore` we can just wait until the lock is released and see if the message was processed successfully or needs to be processed. I have a feeling that this way debugging will be easier since message retry logic will not be involved. WDYT?

kibertoad: Sure, that sounds reasonable. Should we still have some timeout for this?

Drodevbar: In that case probably not, I think. If the lock is kept alive it means that the consumer is doing fine on processing the message and we don't have to add more logic around it.

I'll adjust it today evening 🙇

Drodevbar: @kibertoad Adjusted ✅

packages/redis-message-deduplication-store/lib/RedisMessageDeduplicationStore.ts
Drodevbar: This is no longer the case, since I've adjusted the code to use `redis-semaphore`. Closing it 🙇

README.md
```diff
@@ -101,6 +106,14 @@ Multi-schema consumers support multiple message types via handler configs. They
...
         * `concurrentConsumersAmount` - configuration for specifying the number of concurrent consumers to create. Available only for SQS and SNS consumers
+        * `messageDeduplicationConfig` - configuration for store-based message deduplication on consumer level. For more details on setting this up, see [Consumer-level store-based-message deduplication](#consumer-level-store-based-message-deduplication).
+        * `enableConsumerDeduplication` - enable store-based consumer-level deduplication. For more details on setting this up, see [Consumer-level store-based-message deduplication](#consumer-level-store-based-message-deduplication).
+        * `messageDeduplicationIdField` - which field in the message contains the deduplication id (by default it is `deduplicationId`). This field needs to be defined as `z.string` in the schema. For more details on setting this up, see [Consumer-level store-based-message deduplication](#consumer-level-store-based-message-deduplication).
+        * `messageDeduplicationOptionsField` - which field in the message contains the deduplication options (by default it is `deduplicationOptions`). This field needs to have the below stricture.
```
kibertoad, line 112:
```suggestion
        * `messageDeduplicationOptionsField` - which field in the message contains the deduplication options (by default it is `deduplicationOptions`). This field needs to have the below structure.
```

README.md
```diff
@@ -390,5 +403,167 @@ It needs to implement the following methods:
...
+        // rest of the message payload
+        deduplicationId: 'unique-id',
+        // the below options are optional. In case they are not provided or invalid, default values will be used
+        deduplicationOptions: {
+            deduplicationWindowSeconds: 60,
+        },
+    })
+    // any subsequent call to publish with the same deduplicationId will be ignored for the next 60 seconds
+
+   // You can also publish messages without deduplication, by simply ommitting deduplicationId field
```
kibertoad, line 484:
```suggestion
   // You can also publish messages without deduplication, by simply omitting deduplicationId field
```

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -248,6 +274,11 @@ export abstract class AbstractQueueService<
         ? // @ts-ignore
           message[this.messageTypeField]
         : undefined
+    const messageDeduplicationId =
```
kibertoad, line 277: to reduce probability of errors, if (consumer)publisherDeduplication is enabled, and deduplicationId field is not present, I'd throw an error

kibertoad (approved):

PR #237, opened 2025-01-16
feat(AP-6076): Add consumer-level message deduplication

Drodevbar:
This PR adds consumer-level message deduplication.

I've set the target branch to the one that introduces the publisher-level dedup one as it hasn't been merged yet. We could merge these two at once (I'll adjust versions to bump up only minors) or one by one (then I'll prepare a migration guide and we can keep major version update as in this PR).

Changelog:
1. add consumer-level deduplication mechanism
2. adjust SQS consumer to support it


packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -248,18 +257,13 @@ export abstract class AbstractSqsConsumer<
         // success
         if (result.result) {
           this.handleMessageProcessed(originalMessage, 'consumed')
+          await this.deduplicateMessageAfterProcessing(parsedMessage, true)
```
eduard-fa, line 260: 🟢 The method name sounds a bit misleading. I would expect "de-duplicate" to be something that actually removes duplicates, but it doesn't seem that we are doing anything except just marking the message as consumed(processed).
I would go with either `storeInProgress()` + `markAsCompleted()`
or, to emphasize this is more an exclusive lock obtaining: `acquireLockForProcessing()` + `markAsCompleted()`

packages/sns/lib/sns/AbstractSnsPublisher.ts
```diff
@@ -75,7 +75,7 @@ export abstract class AbstractSnsPublisher<MessagePayloadType extends object>
         () => calculateOutgoingMessageSize(updatedMessage),
       )
 
-      if ((await this.deduplicateMessage(parsedMessage)).isDuplicated) {
+      if ((await this.deduplicateMessageOnPublisherSide(parsedMessage)).isDuplicated) {
```
eduard-fa, line 78: I wonder why the publisher should explicitly emphasize that this is `on publisher side`, we don't have consumer side here 🤔

Drodevbar: This is because both publisher and consumer extend the `AbstracQueueService` class. This is also why you can see many methods on the publisher side marked as `Not implemented for publisher` because they are consumer-specific. There are no generic abstract classes that differentiate one from another. This is why I wanted to emphasize that some of the methods are publisher & consumer specific. I agree it's not the best approach, but I didn't find a better way to do it without breaking the current abstraction.

Drodevbar: How about `deduplicateMessageBeforePublishing`?

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
Drodevbar: Great points, I went with `tryToAcquireLockForProcessing` and `updateLockAfterProcessing`. I've added `tryTo` to the first one as a lock might not always be acquired. I've decided to name the latter one slightly differently to highlight the consumer-related method. `markAsCompleted` sounds quite generic to me while this method is only doing dedup-related work.

Let me know how it sounds to you.

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -525,51 +535,159 @@ export abstract class AbstractQueueService<
...
+    const messageType = message[this.messageTypeField] as string
+
+    return !!this.consumerMessageDeduplicationConfig.messageTypeToConfigMap[messageType]
+  }
+
+  /**
+   * Tries to acquire lock for processing the message.
+   * Returns true in case if lock has been acquired and message can be processed. Returns false otherwise.
+   */
+  protected async tryToAcquireLockForProcessing(message: MessagePayloadSchemas): Promise<boolean> {
```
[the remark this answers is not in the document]
Drodevbar: I was thinking about it but then realized that it would make the solution less generic. The implementation right now is quite generic and relies on the user providing an instance of `ConsumerMessageDeduplicationStore` that serves as a store for message deduplication. MQT provides a package that comes with Redis-based dedup store - `redis-message-deduplication-store`. I think i we would like to use `redis-semphore` we should split this package into 2 - `redis-publisher-message-deduplicaiton-store` (that would rely solely on `ioredis`) and `redis-consumer-message-deduplication-store` (that would rely on both `ioredis` and `redis-sempahore`). How does it sound to you?

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -525,51 +535,159 @@ export abstract class AbstractQueueService<
...
+      return false
+    }
+
+    const deduplicationKeyTtl =
+      await this.consumerMessageDeduplicationConfig.deduplicationStore.getKeyTtl(deduplicationKey)
+
+    // Message got stuck in processing
+    if (
+      deduplicationKeyTtl &&
+      deduplicationKeyTtl <= deduplicationConfig.maximumProcessingTimeSeconds
```
kjamrog, line 636: Probably I'm missing something, but wouldn't we achieve the same behaviour if the key would be set to `deduplicationConfig.maximumProcessingTimeSeconds` and after that time it could be acquired again?

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -525,51 +535,159 @@ export abstract class AbstractQueueService<
...
+
+    if (!deduplicationConfig) {
+      return
+    }
+
+    const deduplicationKey = deduplicationConfig.deduplicationKeyGenerator.generate(message)
+
+    if (messageProcessedSuccessfully) {
+      // Mark the deduplication key as processed and extend its TTL
+      await this.consumerMessageDeduplicationConfig.deduplicationStore.updateKeyTtlAndValue(
```
kjamrog, line 677: 🟢 If the method is named `update`, for me it's not clear what happens if key does not exist. Since there is `setIfNotExists`, I would name this one `set`.
But it might be my personal preference, so let's hear others thoughts as well

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -220,6 +223,13 @@ export abstract class AbstractSqsConsumer<
         }
         const { parsedMessage, originalMessage } = deserializedMessage.result
 
+        if (this.isPublisherDeduplicationEnabled(parsedMessage)) {
```
kjamrog, line 226: Shouldn't it be:
```suggestion
        if (this.isConsumerDeduplicationEnabled(parsedMessage)) {
```
?

Drodevbar: Absolutely. I've fixed it and added sync checks to other methods related to dedup as part of this consumer.

It's weird that the tests were passing. I'll investigate why.

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: I think you're right. This could be simplified. I honestly cannot recall an edge case I was trying to address with such an implementation. I'll rework.

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: There might be an edge case where the key has already expired when we call it. I'll adjust it to support that case as well. Thank you

kjamrog: In that case, the key would be created with status `PROCESSED` right?

Drodevbar: Yes - either created with status PROCESSED or the existing one would be updated (to status PROCESSED) with an extended TTL.

packages/core/lib/queues/AbstractQueueService.ts
Drodevbar: Adjustec ✅

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
Drodevbar: @kjamrog I cannot figure out how the tests are working without the deduplication config. I have some guesses but haven't yet found an exact issue, but to me doesn't mean that there is something wrong with dedup, but rather test config/some bug in the handlerSpy(?)

If you take a look at the test `does not consume message with the same deduplication key twice`, it has 2 checks - the first one checks (using `waitForMessageWithId`) if the message was successfully consumed, the latter one checks (using `checkForMessage`) if the message was not consumed upon subsequent publishing. The tests are passing both with and without the dedup config. Here is the interesting thing. I've noticed that when I replace the second check (`checkForMessage`) with `waitForMessageWithId` then when deduplication is enabled the test fails as a result of a timeout - which is expected and confirms that deduplication works as a consumer cannot consume the message. If I disable deduplication, then test fails, because the message was eventually consumed (which is also expected because there is no dedup configured). So there are different behavior depending on whether `checkForMessage` or `waitForMessage` is used. I suspect that it might be something with the `checkForMessage` behavior, although implementation of this looks simple and I haven't yet figured out what might be an issue there. Is `checkForMessage` actually intended to be called as part of a consumer?

cc @kibertoad

eduard-fa: Could it be that you're not waiting long enough between the second `.publish()` and the `checkForMessage()` call?
The `checkForMessage()` only looks for messages that have already been processed, whereas the message you just published might not have been delivered/handled yet.
Try adding a 1-second delay after the second publish to test this.

To address the issue, we could introduce a new `MessageProcessingResult`, i.e. `rejected` or `duplicate`, to explicitly indicate when a message was rejected because it is a duplicate.
We can then update the expectations to ensure that the `processingResult` is explicitly set to `rejected` or `duplicate` when this occurs.

eduard-fa (commented):
Everything looks good except for the unreliable test you mentioned. Once you address that, I’ll approve it.

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
Drodevbar: @eduard-fa Thanks for a great suggestion. I've introduced a new status `duplicate` and adjusted consumers. Now tests work as expected (if dedup is disabled, then they simply fail).

I'll do the same for consumer to keep it consistent.

eduard-fa (approved):

README.md
```diff
@@ -454,3 +455,64 @@ In such cases, publisher-level deduplication should be combined with consumer-le
         }
     }
     ```
+
+## Consumer-level message deduplication
+
+Consumer-level message deduplication is a mechanism that prevents the same message from being processed multiple times.
+It is useful when you want to be sure that message is processed only once, regardless of how many times it is received.
```
kibertoad, line 462: It would be useful to expand this description. It is unclear what is the mechanism behind the deduplication

packages/core/lib/queues/AbstractQueueService.ts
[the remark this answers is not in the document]
kibertoad: can we move the lock management logic inside the deduplication store then? (primarily move setIfNotExists and getByKey methods inside, and expose tryToAcquireLockForProcessing method from the store interface itself)

packages/redis-message-deduplication-store/lib/RedisConsumerMessageDeduplicationStore.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export type RedisConsumerMessageDeduplicationStoreDependencies = {
+  redis: Redis
+}
+
+export type RedisConsumerMessageDeduplicationStoreConfig = {
+  keyPrefix?: string
+}
+
+export class RedisConsumerMessageDeduplicationStore implements ConsumerMessageDeduplicationStore {
```
kibertoad, line 12: is there a conceptual difference between consumer and publisher deduplication stores? one is not a subset of the other? can we make this class implement two different interfaces (both for publisher and for the consumer)

packages/redis-message-deduplication-store/lib/RedisConsumerMessageDeduplicationStore.ts
```diff
@@ -0,0 +1,60 @@
...
+    return result === 'OK'
+  }
+
+  getByKey(key: string): Promise<string | null> {
+    const keyWithPrefix = this.getKeyWithOptionalPrefix(key)
+
+    return this.redis.get(keyWithPrefix)
+  }
+
+  async getKeyTtl(key: string): Promise<number | null> {
```
kibertoad, line 37: can you add jsdoc, explaining the meaning of null in this case?

packages/redis-message-deduplication-store/lib/RedisConsumerMessageDeduplicationStore.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  constructor(
+    dependencies: RedisConsumerMessageDeduplicationStoreDependencies,
+    config: RedisConsumerMessageDeduplicationStoreConfig,
+  ) {
+    this.redis = dependencies.redis
+    this.config = config
+  }
+
+  async setIfNotExists(key: string, value: string, ttlSeconds: number): Promise<boolean> {
```
kibertoad, line 24: can you add jsdoc, explaining what resulting boolean means? (it's unclear whether it indicates if value existed before, was set, or if operation errored out)

Drodevbar:
Merging this to `feat/AP-6075-producer-level-dedup` as discussed on Slack to ship both producer and consumer dedup together. I'll address all of @kibertoad comments in https://github.com/kibertoad/message-queue-toolkit/pull/236

README.md
Drodevbar: Updated readme as part of https://github.com/kibertoad/message-queue-toolkit/pull/236 - please let me know if the mechanism is clearer for you now 🙇

packages/redis-message-deduplication-store/lib/RedisConsumerMessageDeduplicationStore.ts
Drodevbar: Good point, will do.

Drodevbar: Updated as part of https://github.com/kibertoad/message-queue-toolkit/pull/236

packages/redis-message-deduplication-store/lib/RedisConsumerMessageDeduplicationStore.ts
Drodevbar: Updated as part of https://github.com/kibertoad/message-queue-toolkit/pull/236

PR #238, opened 2025-01-21
Message processing metric improvements

kjamrog:
After analysing first message processing metrics results, the following improvements were identified:

- processing time measurement - currently it operates on message timestamp. It means that if message waits in the queue, or is hold by the barrier, that waiting time is included in the measurement. While this is useful metric, we should also measure processing time itself. 
- Prometheus metric labels:
  - processingResult - most important, that will allow to monitor `published`, `consumed` , `retryLater` statuses separately, currently there are indistinguishable
  - queueName - that will allow us to filter measurements by specific queue
  
See `UPGRADING.md` for the description of changes

### DIFF
--- a/packages/core/lib/types/queueOptionsTypes.ts
+++ b/packages/core/lib/types/queueOptionsTypes.ts
@@ -39,9 +39,20 @@ export type ProcessedMessageMetadata<MessagePayloadSchemas extends object = obje
   queueName: string
 
   /**
-   * Processing time in milliseconds
+   * The timestamp when the message was sent initially, in milliseconds since the epoch
    */
-  messageProcessingMilliseconds?: number
+  messageTimestamp: number | undefined
+
+  /**
+   * The timestamp when the processing of the message begins, in milliseconds since the epoch
+   * Note: for publishers the value may be smaller than messageTimestamp
+   */
+  messageProcessingStartTimestamp: number
+
+  /**
+   * The timestamp when the processing of the message ends, in milliseconds since the epoch
+   */
+  messageProcessingEndTimestamp: number
 }
 
 export interface MessageMetricsManager<MessagePayloadSchemas extends object = object> {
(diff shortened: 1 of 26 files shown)


kibertoad:
I'd say that we need both metrics, actually. Full cycle time is important to understand queue becoming clogged

kjamrog:
@kibertoad yeah, I was thinking about it too. 
What if instead of `messageProcessingMilliseconds`, `MessageMetricsManager` would accept 3 properties:
- `messageTimestamp`
- `messageProcessingStartTimestamp`
- `messageProcessedTimestamp`

And then it would be up to the concrete implementation to register whatever is needed by the client.

`MessageProcessingTimePrometheusMetric` implementation would register 2 metrics, for both measurements.

I think `messageProcessingMilliseconds` name is a better fit for processing time only. And for full cycle time I would create new one then, like: `messageLifetimeMilliseconds`.

kibertoad:
sure, that sounds like a perfect solution!

kjamrog:
Adjusted, `MessageMetricsManager` accepts timestamps now instead of precalculated measurement.

Specific measurements calculations are implemented in metrics classes: `MessageLifetimeMetric` and `MessageProcessingTimeMetric`. Both can be used independently, or combined in `MessageProcessingMultiMetrics` if needed.

CarlosGamero (approved):

packages/core/lib/types/queueOptionsTypes.ts
```diff
@@ -34,9 +34,25 @@ export type ProcessedMessageMetadata<MessagePayloadSchemas extends object = obje
...
-  messageProcessingMilliseconds?: number
+  queueName: string
+
+  /**
+   * The timestamp when the message was sent initially, in milliseconds since the epoch
+   */
+  messageTimestamp: number | undefined
+
+  /**
+   * The timestamp when the processing of the message begins, in milliseconds since the epoch
```
kibertoad, line 47: 🟢 "begins" -> "began", "ends" -> "ended", as we are talking about the actions that have concluded by then.

kibertoad (approved):
great job!

packages/core/lib/types/queueOptionsTypes.ts
kjamrog: Good point, updated

PR #247, opened 2025-03-04
Build fix

CarlosGamero:


packages/amqp/package.json
```diff
@@ -37,7 +37,7 @@
     "@biomejs/biome": "1.9.4",
     "@kibertoad/biome-config": "^1.2.1",
     "@message-queue-toolkit/core": "*",
-    "@types/amqplib": "^0.10.5",
+    "@types/amqplib": "0.10.6",
```
CarlosGamero, line 40: `0.10.7` was released 3 days ago ([doc](https://www.npmjs.com/package/@types/amqplib?activeTab=readme)) and looks like there are some type channges around `Connection`, we need to look into it but pinging prev version to unblock build

kibertoad (approved):

PR #248, opened 2025-03-06
EXP-9 Error metrics

CarlosGamero:


packages/metrics/lib/prometheus/metrics/message-error/PrometheusMessageErrorCounter.ts
```diff
@@ -0,0 +1,37 @@
...
+    this.metric.inc(
+      {
+        queue: metadata.queueName,
+        messageType: metadata.messageType,
+        errorReason: metadata.processingResult.errorReason,
+        version: this.messageVersionGeneratingFunction(metadata),
+      },
+      1,
+    )
+  }
```
CarlosGamero, line 36: Task is about tracking messages which ends up in DLQ, but I think we can extend it to track messages with errors so depending on the service config and the error reason we can know what happen with them, what do you think?

kjamrog (approved):
🚀

kibertoad (commented):
let's update documentation and migration guide as well

kibertoad (approved):
great job!

PR #250, opened 2025-03-10
redis-message-deduplication-store 1.0.0

CarlosGamero:


packages/redis-message-deduplication-store/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@message-queue-toolkit/redis-message-deduplication-store",
-  "version": "1.0.1",
+  "version": "1.0.0",
```
CarlosGamero, line 3: 1.0.0 was not released, so we can release it with the latest changes already

kibertoad (approved):

PR #253, opened 2025-03-21
Add missing handlerSpy for AMQP publisher

kibertoad:
Apparently it was missing all along 😿


packages/amqp/lib/AbstractAmqpPublisher.ts
```diff
@@ -120,6 +129,8 @@ export abstract class AbstractAmqpPublisher<
     }
   }
 
+  protected abstract resolveTopicOrQueue(): string
```
kibertoad, line 132: I'm very much not a fan of this solution, but I don't know what would be a better way. we know whether we want to pass queue or topic within more specific publisher class, but they no longer have the non-serialized message

matfin (approved):
🟢 Looks good to me

PR #255, opened 2025-04-18
Added queueName to SQS locator config

kjamrog:
Providing `queueUrl` in the locator config is not always convenient, as it requires users to provide AWS connection details. This PR introduces `queueName` parameter for locator config, that can be provided instead. The corresponding queueUrl is then resolved dynamically.

### DIFF
--- a/packages/sns/package.json
+++ b/packages/sns/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@message-queue-toolkit/sns",
-  "version": "21.0.1",
+  "version": "21.1.0",
   "private": false,
   "license": "MIT",
   "description": "SNS adapter for message-queue-toolkit",
--- a/packages/sns/package.json
+++ b/packages/sns/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@message-queue-toolkit/sns",
-  "version": "21.0.0",
+  "version": "21.0.1",
   "private": false,
   "license": "MIT",
   "description": "SNS adapter for message-queue-toolkit",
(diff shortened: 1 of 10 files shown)


packages/sqs/lib/sqs/AbstractSqsService.ts
```diff
@@ -24,7 +24,8 @@ export type SQSCreationConfig = {
 } & ExtraSQSCreationParams
 
 export type SQSQueueLocatorType = {
-  queueUrl: string
+  queueUrl?: string
+  queueName?: string
 }
```
CarlosGamero, line 29: Nit: should this type be allow only either `queueName` or `queueUrl`

CarlosGamero (approved):
🙏🚀

packages/sqs/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@message-queue-toolkit/sqs",
-  "version": "20.0.0",
+  "version": "20.0.1",
```
CarlosGamero, line 3: I would say this is a minor release

packages/sqs/lib/sqs/AbstractSqsService.ts
kjamrog: Adjusted

packages/sqs/package.json
kjamrog: Agree, updated ✅

packages/sns/package.json
```diff
@@ -35,7 +35,7 @@
     "@aws-sdk/client-sts": "^3.632.0",
     "@message-queue-toolkit/core": ">=20.0.0",
     "@message-queue-toolkit/schemas": ">=2.0.0",
-    "@message-queue-toolkit/sqs": ">=20.0.0"
+    "@message-queue-toolkit/sqs": ">=20.0.1"
```
CarlosGamero, line 38:
```suggestion
    "@message-queue-toolkit/sqs": ">=20.1.0"
```

CarlosGamero (approved):

packages/sns/package.json
kjamrog: Great catch, thanks 🙏 already updated

kibertoad (approved):

PR #259, opened 2025-04-25
Automatically init non-lazy publishers on PublisherManager creation

ForterLokalise:
After introducing option to disable lazy init for queues, there's no way to actually initialise them.
This PR introduces automatic initialisation of registered queues when lazy init is disabled, at the moment when PublisherManager is created.

### DIFF
--- a/packages/amqp/README.md
+++ b/packages/amqp/README.md
@@ -71,4 +71,4 @@ publisher.publish(message)
 
 By default `message-queue-toolkit/amqp` will lazily instantiate publishers before the publish in case they weren't instantiated before. This may result in system not failing fast when connection to AMQP cannot be established, as instantiation happens asynchronously. In case that is an undesirable behaviour, you can set parameter `isLazyInitEnabled` to false either directly for a publisher, or as one of the `newPublisherOptions` parameters for the `AmqpQueuePublisherManager`.
 
-In case when lazy instantiation is disabled, queues should be initilized explicitly by using an async `AmqpQueuePublisherManager::initRegisteredQueues()` method.
+In case when lazy instantiation is disabled, queues should be initilized explicitly by using an async `AmqpQueuePublisherManager.initRegisteredQueues()` method.
--- a/packages/amqp/README.md
+++ b/packages/amqp/README.md
@@ -71,4 +71,4 @@ publisher.publish(message)
 
 By default `message-queue-toolkit/amqp` will lazily instantiate publishers before the publish in case they weren't instantiated before. This may result in system not failing fast when connection to AMQP cannot be established, as instantiation happens asynchronously. In case that is an undesirable behaviour, you can set parameter `isLazyInitEnabled` to false either directly for a publisher, or as one of the `newPublisherOptions` parameters for the `AmqpQueuePublisherManager`.
 
-In case when lazy instantiation is disabled, publishers should be initilized explicitly by using an async `AmqpQueuePublisherManager.initRegisteredPublishers()` method.
+In case when lazy instantiation is disabled, publishers should be initilized explicitly by using an async `AmqpQueuePublisherManager.initRegisteredPublishers()` method. It also has an optional `publishers?: string[]` argument, which can be used to only initialize some specific publishers.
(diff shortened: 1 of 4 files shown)


packages/amqp/lib/AmqpQueuePublisherManager.ts
```diff
@@ -132,6 +136,16 @@ export class AmqpQueuePublisherManager<
     }
   }
 
+  protected initRegisteredQueues(): void {
+    for (const eventTarget in this.targetToEventMap) {
+      const queueName = eventTarget as NonNullable<SupportedEventDefinitions[number]['queueName']>
+
+      if (this.targetToPublisherMap[queueName]) {
+        this.targetToPublisherMap[queueName].init()
```
kibertoad, line 144: isn't this async? we can't call this in constructor, it has to be a separate async method.

packages/amqp/lib/AmqpQueuePublisherManager.ts
```diff
@@ -132,6 +136,16 @@ export class AmqpQueuePublisherManager<
     }
   }
 
+  protected initRegisteredQueues(): void {
```
kibertoad, line 139: test is missing for this functionality

packages/amqp/lib/AmqpQueuePublisherManager.ts
ForterLokalise: Agree, it will be called by client code. Adjusted in https://github.com/kibertoad/message-queue-toolkit/pull/259/commits/a35a17ac4296f177c6d56f708644f68eb114117b

packages/amqp/lib/AmqpQueuePublisherManager.ts
```diff
@@ -132,6 +132,14 @@ export class AmqpQueuePublisherManager<
     }
   }
 
+  async initRegisteredQueues(): Promise<void> {
```
kibertoad, line 135: can you also add this method to the readme?

packages/amqp/lib/AmqpQueuePublisherManager.ts
ForterLokalise: Implemented in https://github.com/kibertoad/message-queue-toolkit/pull/259/commits/fedb701b1880316c1b27739ecebd9209c0523d0f

packages/amqp/lib/AmqpQueuePublisherManager.ts
ForterLokalise: Added in https://github.com/kibertoad/message-queue-toolkit/pull/259/commits/d95982b5d4a028b7944a9cd0e4add66dbf0378d7

packages/amqp/README.md
```diff
@@ -70,3 +70,5 @@ publisher.publish(message)
 ## Lazy instantiation
 
 By default `message-queue-toolkit/amqp` will lazily instantiate publishers before the publish in case they weren't instantiated before. This may result in system not failing fast when connection to AMQP cannot be established, as instantiation happens asynchronously. In case that is an undesirable behaviour, you can set parameter `isLazyInitEnabled` to false either directly for a publisher, or as one of the `newPublisherOptions` parameters for the `AmqpQueuePublisherManager`.
+
+In case when lazy instantiation is disabled, queues should be initilized explicitly by using an async `AmqpQueuePublisherManager::initRegisteredQueues()` method.
```
kibertoad, line 74:
```suggestion
In case when lazy instantiation is disabled, queues should be initilized explicitly by using an async `AmqpQueuePublisherManager.initRegisteredQueues()` method.
```

packages/amqp/lib/AmqpQueuePublisherManager.ts
kibertoad, line 135: can you also add an optional string[] param to make it possible to only init publishers for specified topics? (so that we could avoid bootstrappung publishers which we do not need, which may be expensive due to creating topic etc)

ForterLokalise: Done in https://github.com/kibertoad/message-queue-toolkit/pull/259/commits/6f9a46e66171f41cc4fce9aa6fb2357d9ec5702f

kibertoad (approved):
thanks a lot!

PR #260, opened 2025-04-28
Chore: TECH-894 kafka bootstrap

CarlosGamero:


packages/kafka/package.json
```diff
@@ -0,0 +1,61 @@
...
+  "version": "0.0.0",
+  "private": false,
+  "license": "MIT",
+  "homepage": "https://github.com/kibertoad/message-queue-toolkit",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/kibertoad/message-queue-toolkit.git"
+  },
+  "description": "Kafka adapter for message-queue-toolkit",
+  "keywords": ["message", "queue", "queues", "abstract", "common", "utils", "notification", "sns"],
```
kibertoad, line 12:
```suggestion
  "keywords": ["message", "queue", "queues", "abstract", "common", "utils", "notification", "kafka"],
```

packages/kafka/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "test": "vitest",
+    "test:coverage": "npm test -- --coverage",
+    "test:ci": "npm run docker:start:dev && npm run test:coverage && npm run docker:stop:dev",
+    "lint": "biome check . && tsc",
+    "lint:fix": "biome check --write .",
+    "docker:start:dev": "docker compose up -d",
+    "docker:stop:dev": "docker compose down",
+    "prepublishOnly": "npm run build:release"
+  },
+  "dependencies": {
```
kibertoad, line 39: can you also add https://www.npmjs.com/package/node-rdkafka as a dependency and ensure it installs correctly in CI? (since it is a native library, I imagine it needs to be built and likely is picky about node versions it supports for that)

CarlosGamero: yeah, will add it later or tomorrow

packages/kafka/package.json
CarlosGamero: Ups, my bad 😓, great catch! thanks Igor 🙏

packages/kafka/package.json
CarlosGamero: Done, also added some stupid tests to make sure it works but still trying to make them reliable ... the produce/consume one is not always working (actually is failing more than working) not sure why

packages/kafka/lib/test.spec.ts
```diff
@@ -0,0 +1,67 @@
+import { once } from 'node:events'
+import { waitAndRetry } from '@lokalise/universal-ts-utils/node'
+import { KafkaConsumer, type Message, Producer, features, librdkafkaVersion } from 'node-rdkafka'
+
+// TODO: to be removed once we have proper tests
+describe('Test', () => {
+  it('should use node-rdkafka', () => {
+    expect(features).toBeDefined()
+    expect(librdkafkaVersion).toMatchInlineSnapshot(`"2.8.0"`)
```
kibertoad, line 9: this is pretty fragile, let's just assert it is defined

CarlosGamero:
Finally green! 🚀 
I will test https://blog.platformatic.dev/why-we-created-another-kafka-client-for-nodejs?showSharer=true in a follow-up PR

docker-compose.yml
```diff
@@ -34,6 +37,13 @@ services:
       - redis_data:/redis/data:cached
     restart: on-failure
 
+  kafka:
+    image: apache/kafka:latest
```
kibertoad, line 41: it's generally recommended to pin infra dependencies to specific versions to ensure runs are deterministicc. let's use the latest one for now, and adjust after we have actual infra

kibertoad (approved):

PR #261, opened 2025-04-28
Chore: ESM migration

CarlosGamero:


packages/amqp/package.json
```diff
@@ -10,40 +10,45 @@
...
-    "@message-queue-toolkit/core": ">=20.0.0",
-    "@message-queue-toolkit/schemas": ">=2.0.0",
+    "@message-queue-toolkit/core": ">=21.0.0",
+    "@message-queue-toolkit/schemas": ">=6.0.0",
     "amqplib": "^0.10.5"
   },
   "devDependencies": {
     "@biomejs/biome": "1.9.4",
     "@kibertoad/biome-config": "^1.2.1",
+    "@lokalise/tsconfig": "^1.3.0",
```
kibertoad, line 43: I wonder if we should switch to lokalise biome config as well, any thoughts?

CarlosGamero: I had the same question, and was going to ask you tomorrow (today we had a massive connection and electricity issue in Spain, service is starting to recover just now).

If you agree with using lokalise/biome, I would make it independently, those changes should not require a release

kibertoad: yeah, I think it would be a good fit

CarlosGamero: Ok! adding it to my todo list :D

CarlosGamero: At the and I had to it in this same PR ... 😓 we need `.js/.ts` imports in ESM and migration to lokalise/biome is easier than reverting tsconfig one

CarlosGamero: Sorry for the huge PR 🙏 please ignore all code changes, they are only about lint fixes and imports.

important changes are `package.json`  `tsconfig` and `vitest.config.ts`

kibertoad (approved):

PR #263, opened 2025-05-02
Chore: platformatic kafka

CarlosGamero:


kibertoad (approved):

packages/kafka/package.json
```diff
@@ -44,14 +44,14 @@
         "prepublishOnly": "npm run lint && npm run build"
     },
     "dependencies": {
```
kibertoad, line 46: let's also add engine entry, matching the platformatic kafka one, to clearly communicate requirements, and also probably mention it in the readme

packages/kafka/test/testContext.ts
```diff
@@ -1,17 +1,24 @@
-/*
 import type { CommonLogger } from '@lokalise/node-core'
+import { type AwilixContainer, Lifetime, type Resolver, asFunction, createContainer } from 'awilix'
+import { AwilixManager } from 'awilix-manager'
 
-export const SINGLETON_CONFIG = { lifetime: Lifetime.SINGLETON }
+const SINGLETON_CONFIG = { lifetime: Lifetime.SINGLETON }
+
+type DiConfig = Record<keyof Dependencies, Resolver<any>>
```
kibertoad, line 7: I think the type that we use now is `type DiConfig = NameAndRegistrationPair<Dependencies>`

kibertoad (approved):

packages/kafka/test/testContext.ts
CarlosGamero: Good catch! already updated :D

PR #268, opened 2025-05-02
Chore: CI adjusting docker compose

CarlosGamero:
Adjusting the code to have a single docker-compose and allow starting only the needed containers with the new CI action


CarlosGamero:
Screenshot 2025-05-02 at 23 38 24

packages/amqp/package.json
```diff
@@ -20,12 +20,11 @@
         "build": "npm run clean && tsc --project tsconfig.build.json",
         "clean": "rimraf dist",
         "test": "vitest",
-        "test:coverage": "npm test -- --coverage",
-        "test:ci": "npm run docker:start:dev && npm run test:coverage && npm run docker:stop:dev",
+        "test:coverage": "npm run test -- --coverage",
```
kibertoad, line 23: I would remove the non-coverage one altogether

CarlosGamero: I reached out in Slack to make a decission about this, in order to proceed with my current I will merge this PR as it is, and if we decide to remove the non-coverage command, I will address it in follow-up PRs

PR #276, opened 2025-05-14
Chore/node core 14

CarlosGamero:


.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   general:
     strategy:
       matrix:
-        node-version: [18.x, 20.x, 22.x, 24.x]
+        node-version: [20.x, 22.x, 24.x]
```
CarlosGamero, line 13: @kibertoad including this here already to avoid creating a new PR just for this

CatchMe2 (approved):

packages/sqs/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@message-queue-toolkit/sqs",
-    "version": "21.0.0",
+    "version": "22.0.0",
```
kibertoad, line 3: should this be a breaking change? does anything change for the end users?

CatchMe2: If any custom error is exported by the lib, then yeah, the narrowed types could by breaking

CarlosGamero: I had the same question, so I asked AI, and the answer is that:
Even without changes on the public API, this change can make clients update their node-core versions, so ideally it should be a breaking change to inform about it

CarlosGamero: I do not have stron opinion, so if you think patch is better here, I am happy to change it

kibertoad (approved):

PR #279, opened 2025-05-22
Confluentinc kafka javascript

CarlosGamero:


docker-compose.yml
```diff
@@ -42,6 +42,33 @@ services:
...
+    image: provectuslabs/kafka-ui:latest
+    ports:
+      - 8080:8080
+    depends_on:
+      - kafka
+    environment:
+      DYNAMIC_CONFIG_ENABLED: 'true'
+      KAFKA_CLUSTERS_0_NAME: Local
+      KAFKA_CLUSTERS_0_BOOTSTRAPSERVERS: kafka:9093
     restart: on-failure
```
CarlosGamero, line 72: This is useful to debug tests 😅

CarlosGamero:
Kafka CI is failing due to `npm install --ignore-scripts`, we need to execute scripts for `@confluentinc/kafka-javascript`, if we decide to go with this library I will address it :D

CarlosGamero:
We have developed a similar test for the libraries under consideratio, it can be used to get an initial sense of their performance. Please take these numbers with a grain of salt—they do not represent formal benchmarks, and more effort would be required for accurate benchmarking.

Performance results from our preliminary tests:

- [node-rdkafka](https://github.com/kibertoad/message-queue-toolkit/pull/281)
  - Test run time: 4s
  - Script build time: 1m 27s
  - Does not support threads
- [confluentinc-kafka](https://github.com/kibertoad/message-queue-toolkit/pull/282)
  - Test run time: 3s
  - Script build time: 1s
  - Does not support threads
- [platformatic-kafka](https://github.com/kibertoad/message-queue-toolkit/pull/280)
  - Test run time: 3s
  - No script build step required
  - Supports threads

packages/kafka/lib/test.spec.ts
```diff
@@ -1,17 +1,8 @@
...
+import { KafkaJS } from '@confluentinc/kafka-javascript'
 import { waitAndRetry } from '@lokalise/universal-ts-utils/node'
-import {
-  Consumer,
-  type Message,
-  Producer,
-  stringDeserializers,
-  stringSerializers,
-} from '@platformatic/kafka'
-import { ProduceAcks } from '@platformatic/kafka'
```
kibertoad, line 10: what do you think about keeping all three engines for now?

kibertoad: or two most promising at least, node-rdkafka so far seems the least useful

packages/kafka/lib/test.spec.ts
```diff
@@ -23,51 +22,112 @@ describe('Test', () => {
...
-    // When
-    await producer.send({
-      messages: [{ topic, value: messageValue }],
-      acks: ProduceAcks.NO_RESPONSE,
+      expect(messages).toHaveLength(1)
+      expect(messages[0]).toEqual(messageValue)
     })
+  })
+
+  describe('platformatic', () => {
```
kibertoad, line 83: let's maybe move to a separate file?

packages/kafka/package.json
```diff
@@ -1,7 +1,9 @@
 {
     "name": "@message-queue-toolkit/kafka",
     "version": "0.0.0",
-    "engines": { "node": ">= 22.14.0" },
+    "engines": {
+        "node": ">= 20.0.0"
```
kibertoad, line 5: I assume we need 22.14.0 again now?

CarlosGamero: Oh, you are right! 😓

kibertoad (approved):

PR #285, opened 2025-06-03
Feat: kafka consumer

CarlosGamero:


packages/kafka/lib/handler-container/KafkaHandlerContainer.ts
```diff
@@ -0,0 +1,66 @@
...
+  TopicConfig,
+} from '../types.ts'
+import type { KafkaHandlerConfig } from './KafkaHandlerConfig.ts'
+import type { KafkaHandlerRouting } from './KafkaHandlerRoutingBuilder.ts'
+
+const DEFAULT_HANDLER_KEY = Symbol('default-handler')
+
+type Handlers<TopicsConfig extends TopicConfig[]> = Record<
+  string,
+  Record<string | symbol, KafkaHandlerConfig<SupportedMessageValues<TopicsConfig>>>
```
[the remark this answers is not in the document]
CarlosGamero: I tried it, but it will force us to always have the `DEFAULT_HANDLER_KEY` key in the record, not sure why 😓, but as it is an internal type and class, I think we do not need to narrow down types here

packages/kafka/lib/AbstractKafkaConsumer.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  async init(): Promise<void> {
+    if (this.consumerStream) return Promise.resolve()
+
+    this.consumerStream = await this.consumer.consume({
+      ...this.options,
+      topics: this.handlerContainer.topics,
+    })
+
+    this.consumerStream.on('data', async (message) => {
```
kibertoad, line 62: do we also need a listener for errors?

packages/kafka/lib/handler-container/KafkaHandlerContainer.ts
```diff
@@ -0,0 +1,70 @@
...
+  ): Handlers<TopicsConfig> {
+    const result: Handlers<TopicsConfig> = {}
+
+    for (const [topic, topicHandlers] of Object.entries(topicHandlerRouting)) {
+      if (!topicHandlers.length) continue
+      result[topic] = {}
+
+      for (const handler of topicHandlers) {
+        let handlerKey = this.messageTypeField
+          ? // @ts-ignore
```
kibertoad, line 37: let's do ts-expect-error maybe?

packages/kafka/lib/AbstractKafkaConsumer.ts
CarlosGamero: Mmmm probably 🤔, I will check if we need to listen for those events

packages/kafka/lib/handler-container/KafkaHandlerContainer.ts
CarlosGamero: Good point! will change it :D

packages/kafka/lib/AbstractKafkaConsumer.ts
```diff
@@ -0,0 +1,157 @@
...
+export type KafkaConsumerOptions<TopicsConfig extends TopicConfig[]> = BaseKafkaOptions &
+  Omit<
+    ConsumerOptions<string, object, string, object>,
+    'deserializers' | 'autocommit' | 'bootstrapBrokers'
+  > &
+  Omit<ConsumeOptions<string, object, string, object>, 'topics'> & {
+    handlers: KafkaHandlerRouting<TopicsConfig>
+  }
+
+const MAX_IN_MEMORY_RETRIES = 3
```
CarlosGamero, line 27: Discussed on daily -> in the first version, we have in-memory retries, and later we will implement a proper retry mechanism + DLQ

packages/core/lib/queues/MessageSchemaContainer.ts
```diff
@@ -8,75 +8,59 @@ export type MessageSchemaContainerOptions<MessagePayloadSchemas extends object>
   messageTypeField?: string
 }
 
-const NO_MESSAGE_TYPE = 'NO_MESSAGE_TYPE'
+const DEFAULT_SCHEMA_KEY = Symbol('NO_MESSAGE_TYPE')
```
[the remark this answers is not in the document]
CarlosGamero: Yeah, the idea is to keep it strictly internal, actually, we are not even exposing handler containers

packages/kafka/lib/AbstractKafkaConsumer.ts
kibertoad: dowe have a ticket for it?

does kafka have a concept of dlq?

packages/kafka/test/consumer/PermissionConsumer.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+describe('PermissionConsumer', () => {
+  let testContext: TestContext
+  let consumer: PermissionConsumer | undefined
+
+  beforeAll(async () => {
+    testContext = await createTestContext()
+  })
+
+  afterEach(async () => {
+    await consumer?.close()
```
kibertoad, line 19: why do we need this? we don't seem to restart it after

packages/kafka/test/consumer/PermissionConsumer.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+      await consumer.init()
+
+      // Then
+      await expect(consumer.init()).resolves.not.toThrowError()
+    })
+
+    it('should fail if kafka is not available', async () => {
+      // Given
+      consumer = new PermissionConsumer(testContext.cradle, {
+        kafka: { clientId: randomUUID(), bootstrapBrokers: ['localhost:9091'] },
```
kibertoad, line 77: should the address be hardcoded. and inlined?

kibertoad (approved):
great job!

packages/kafka/lib/AbstractKafkaConsumer.ts
CarlosGamero: We do not have it yet, but I am going to create it and link it here :D 

> does kafka have a concept of dlq?

Not really, but we can create a topic with `-dlq` suffix and send there all failed messages 🤔. In any case, we need to discuss and design it properly 😅

packages/kafka/test/consumer/PermissionConsumer.spec.ts
CarlosGamero: We are starting the consumer on each test with different parameters. We need to close it after each test to avoid collisions (even though we are using different group IDs, better to be safe)

packages/kafka/lib/AbstractKafkaConsumer.ts
CarlosGamero: https://lokalise.atlassian.net/browse/EDEXP-498

packages/kafka/test/consumer/PermissionConsumer.spec.ts
CarlosGamero: In this test we are checking if the consumer fails if Kafka does not exist. This address (port) is wrong, but I see it can confuse, going to change it to something more meaningful

PR #297, opened 2025-06-27
Kafka: handle commit errors and add logging

kjamrog:
Summary of changes:
- some errors on offset commit are expected and related to group rebalancing when new consumer joins. This PR adds graceful handling for them - we are logging them but they don't need to be thrown
- added event listeners with logs for better observability of the consumer lifetime


packages/amqp/package.json
```diff
@@ -37,7 +37,7 @@
         "amqplib": "^0.10.8"
     },
     "devDependencies": {
-        "@biomejs/biome": "2.0.5",
+        "@biomejs/biome": "^1.9.4",
```
kjamrog, line 40: Switching back to biome `1.9.4` as the new version is not compatible with our config in `lokalise/biome-config`. We need to upgrade there first. (The version in MQT was updated automatically few days ago, that's why we didn't have issues before: https://github.com/kibertoad/message-queue-toolkit/pull/296/files)

packages/kafka/lib/AbstractKafkaConsumer.ts
```diff
@@ -192,6 +209,48 @@ export abstract class AbstractKafkaConsumer<
...
 
+  private async commitMessage(message: Message<string, object, string, string>) {
+    try {
+      this.logger.debug(
+        { topic: message.topic, offset: message.offset, timestamp: message.timestamp },
+        'Trying to commit message',
+      )
+      await message.commit()
+    } catch (error) {
+      if (error instanceof ResponseError && 'code' in error) {
```
kibertoad, line 220: why the code condition?

packages/kafka/lib/AbstractKafkaConsumer.ts
```diff
@@ -192,6 +209,48 @@ export abstract class AbstractKafkaConsumer<
...
+    }
+  }
+
+  private handleResponseErrorOnCommit(responseError: ResponseError) {
+    // Some errors are expected during group rebalancing, so we handle them gracefully
+    for (const error of responseError.errors) {
+      if (
+        error instanceof ProtocolError &&
+        error.apiCode &&
+        (error.apiCode === ILLEGAL_GENERATION ||
```
kibertoad, line 233: wonder if it could be an set member check

packages/kafka/lib/AbstractKafkaConsumer.ts
```diff
@@ -90,11 +104,6 @@ export abstract class AbstractKafkaConsumer<
 
     try {
       const { handlers, ...consumeOptions } = this.options // Handlers cannot be passed to consume method
-      await this.consumer.joinGroup({
-        sessionTimeout: consumeOptions.sessionTimeout,
-        rebalanceTimeout: consumeOptions.rebalanceTimeout,
-        heartbeatInterval: consumeOptions.heartbeatInterval,
-      })
```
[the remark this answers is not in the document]
CarlosGamero: Mmm, I see, I will add it again then :D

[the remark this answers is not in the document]
CarlosGamero: Thank youuu

PR #298, opened 2025-06-29
Migrate to zod v4

kibertoad:


.github/workflows/ci.common.yml
```diff
@@ -39,3 +39,6 @@ jobs:
 
       - name: Docker stop
         run: npm run docker:stop:ci -- --filter=${{ inputs.package_name }}
+
+      - name: Run lint
```
kibertoad, line 43: previously automerge would still happen if linting failed, now this shouldn't be the case

packages/kafka/package.json
```diff
@@ -54,15 +54,15 @@
         "zod": "^3.25.7"
     },
     "peerDependencies": {
-        "@message-queue-toolkit/core": ">=21.3.0",
-        "@message-queue-toolkit/schemas": ">=6.0.0"
+        "@message-queue-toolkit/core": "21.3.0",
+        "@message-queue-toolkit/schemas": "6.2.0"
```
kibertoad, line 58: pinned for now, let's fix Kafka in a separate PR, it's far from trivial

packages/amqp/package.json
```diff
@@ -29,15 +29,15 @@
...
-        "zod": "^3.25.7"
+        "zod": "^3.25.67"
     },
     "peerDependencies": {
         "@message-queue-toolkit/core": ">=21.1.0",
         "@message-queue-toolkit/schemas": ">=6.0.0",
         "amqplib": "^0.10.8"
     },
     "devDependencies": {
-        "@biomejs/biome": "2.0.5",
```
kamilwylegala, line 40: Why reverting biome?

packages/sns/lib/types/MessageTypes.ts
```diff
@@ -14,7 +14,7 @@ export const SNS_MESSAGE_BODY_SCHEMA = z.object({
       z.string(),
       z.object({
         Type: z.string(),
-        Value: z.unknown(),
+        Value: z.any(),
```
kamilwylegala, line 17: It's new Zod's unknown or just `any`?

kibertoad: their type inference changed, previously z.unknown would be inferred as any, now it is more strict and becomes unknown, which end up resolving entire object type as unknown in several cases, so in order to preserve original behaviour, I've did the change. we can experiment with restoring unknown and making types stricter separately from this PR

packages/schemas/lib/messages/baseMessageSchemas.ts
```diff
@@ -38,27 +40,34 @@ export type MetadataObject = ZodObject<
     originatedFrom: ZodString
     correlationId: ZodString
   },
-  'strip'
+  core.$strip
```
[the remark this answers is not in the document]
kibertoad: lemme try remove it

PR #429, opened 2026-04-28
fix(core): preserve type through offloading + reject removed messageTypeField

kibertoad:
## Summary

Two related fixes for the same regression class — silent drop of the message \`type\` field through payload offloading.

### 1. Type-level rejection of the removed `messageTypeField` option

`messageTypeField` was removed in core 25.x in favor of `messageTypeResolver`. Until now the option was simply absent from `CommonQueueOptions`, but TypeScript's excess-property checks are weakened across the long `Omit`/`&`/generic chains used by publisher options (e.g. `SnsPublisherManager.newPublisherOptions`), so callers who hadn't migrated kept silently passing a no-op option.

Fix: type \`messageTypeField\` as a removal-marker string-literal (\`'__REMOVED_USE_messageTypeResolver_INSTEAD__'\`). Passing it now fails compilation with a diagnostic that surfaces the migration hint at the call site.

### 2. Consistent type-field preservation through offloading

`AbstractQueueService.offloadMessagePayloadIfNeeded` preserved the `type` field only when `messageTypeResolver` was the `messageTypePath` flavor. With no resolver, or with `literal` / `resolver` modes, the field was silently stripped. This was inconsistent with:

- the rest of the offloader, which copies `messageIdField` / `messageTimestampField` / `messageDeduplicationIdField` / `messageDeduplicationOptionsField` unconditionally based on their **defaulted** field names (\`'id'\`, \`'timestamp'\`, ...);
- `MessageSchemaContainer.resolveSchema`, which falls back to a single default schema when no resolver is configured.

Fix: default the type-preservation path to the conventional top-level \`type\` when the resolver doesn't provide a body path. The literal/resolver modes still drive routing/dispatch as before; this only keeps the field present for downstream consumers (notably SNS subscriptions whose FilterPolicy filters on the body \`type\`).

Production-confirmed regression in file-storage-service: after the 24.x → 25.x bump, large \`statistic.generated\` messages were getting offloaded successfully and SNS-published successfully, but never reached SQS because the SQS subscription's \`FilterPolicyScope: 'MessageBody'\` filter on \`type\` no longer matched. See lokalise/file-storage-service#1027.

## Test plan

- [x] \`tsc --noEmit\` clean in \`packages/core\` and \`packages/sns\` after the change.
- [x] New unit tests in \`packages/core/test/queues/AbstractQueueService.offload.spec.ts\` cover all four resolver modes (none / messageTypePath / literal / resolver), the nested path case, and the no-\`type\` source case. Verified that 3 of these fail without the offload fix and all pass with it.
- [x] New type-level tests in \`packages/core/test/types/queueOptionsTypes.types.spec.ts\` and \`packages/sns/test/SnsPublisherManager.types.spec.ts\` lock in compile-time rejection of the legacy option.
- [x] Full core test suite (\`vitest run packages/core\`): 168 passed.

## Summary by CodeRabbit

* **Documentation**
  * Deprecated the legacy messageTypeField option; use messageTypeResolver instead (messageTypePath, literal, or resolver).

* **Bug Fixes**
  * Offloaded message payloads now preserve the original message type when present, preventing downstream filter mismatches.

* **Tests**
  * Added compile-time type tests to enforce removal of the deprecated option, regression tests to verify offloaded payloads retain type, and enabled type-checking for type-level tests.

### DIFF
--- a/packages/core/lib/types/queueOptionsTypes.ts
+++ b/packages/core/lib/types/queueOptionsTypes.ts
@@ -119,10 +119,10 @@ export type CommonQueueOptions = {
    * (or `{ literal: '<type>' }` / `{ resolver: fn }`) instead. See UPGRADING.md.
    *
    * Typed as a removal-marker string literal so callers passing the legacy option get a
-   * compile-time error whose diagnostic shows the migration hint, rather than a silent
-   * runtime drop. Leaving this set on a publisher with payload offloading causes the
-   * message `type` field to be stripped from the offloaded SNS body, which then silently
-   * fails any downstream subscription whose FilterPolicy filters on `type`.
+   * compile-time error whose diagnostic shows the migration hint at the call site, even
+   * across long `Omit`/`&`/generic chains where excess-property check no longer fires
+   * (the realistic regression: `messageTypeField: '<name>'` flowing through e.g. the
+   * `SnsPublisherManager.newPublisherOptions` chain).
    *
    * Note: passing `messageTypeField: undefined` still type-checks (without
    * `exactOptionalPropertyTypes`), but `undefined` is operationally identical to omitting
--- a/packages/core/lib/types/queueOptionsTypes.ts
+++ b/packages/core/lib/types/queueOptionsTypes.ts
@@ -118,12 +118,17 @@ export type CommonQueueOptions = {
    * @deprecated Removed in core 25.x. Use `messageTypeResolver: { messageTypePath: '<field>' }`
    * (or `{ literal: '<type>' }` / `{ resolver: fn }`) instead. See UPGRADING.md.
    *
-   * Typed as `never` so callers that still pass it get a compile-time error rather than
-   * a silent runtime drop. Notably, leaving this set on a publisher with payload offloading
-   * causes the message `type` field to be stripped from the offloaded SNS body, which then
-   * silently fails any downstream subscription whose FilterPolicy filters on `type`.
+   * Typed as a removal-marker string literal so callers passing the legacy option get a
+   * compile-time error whose diagnostic shows the migration hint, rather than a silent
+   * runtime drop. Leaving this set on a publisher with payload offloading causes the
+   * message `type` field to be stripped from the offloaded SNS body, which then silently
+   * fails any downstream subscription whose FilterPolicy filters on `type`.
+   *
+   * Note: passing `messageTypeField: undefined` still type-checks (without
+   * `exactOptionalPropertyTypes`), but `undefined` is operationally identical to omitting
+   * the field, so this guard catches the realistic regression (`messageTypeField: '<name>'`).
    */
-  messageTypeField?: never
+  messageTypeField?: '__REMOVED_USE_messageTypeResolver_INSTEAD__'
   messageIdField?: string
   messageTimestampField?: string
   messageDeduplicationIdField?: string
--- a/packages/core/lib/types/queueOptionsTypes.ts
+++ b/packages/core/lib/types/queueOptionsTypes.ts
@@ -114,6 +114,16 @@ export type CommonQueueOptions = {
    * }
    */
   messageTypeResolver?: MessageTypeResolverConfig
+  /**
+   * @deprecated Removed in core 25.x. Use `messageTypeResolver: { messageTypePath: '<field>' }`
+   * (or `{ literal: '<type>' }` / `{ resolver: fn }`) instead. See UPGRADING.md.
+   *
+   * Typed as `never` so callers that still pass it get a compile-time error rather than
+   * a silent runtime drop. Notably, leaving this set on a publisher with payload offloading
+   * causes the message `type` field to be stripped from the offloaded SNS body, which then
+   * silently fails any downstream subscription whose FilterPolicy filters on `type`.
+   */
+  messageTypeField?: never
   messageIdField?: string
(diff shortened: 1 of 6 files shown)


packages/core/lib/types/queueOptionsTypes.ts
```diff
@@ -114,6 +114,16 @@ export type CommonQueueOptions = {
...
+  /**
+   * @deprecated Removed in core 25.x. Use `messageTypeResolver: { messageTypePath: '<field>' }`
+   * (or `{ literal: '<type>' }` / `{ resolver: fn }`) instead. See UPGRADING.md.
+   *
+   * Typed as `never` so callers that still pass it get a compile-time error rather than
+   * a silent runtime drop. Notably, leaving this set on a publisher with payload offloading
+   * causes the message `type` field to be stripped from the offloaded SNS body, which then
+   * silently fails any downstream subscription whose FilterPolicy filters on `type`.
+   */
+  messageTypeField?: never
```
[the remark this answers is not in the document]
kibertoad: Good catch on the `undefined` loophole. Two practical considerations:

1. **Closing it requires enabling `exactOptionalPropertyTypes` package-wide.** I tried that locally — it produces ~13 unrelated errors in `lib/queues/AbstractQueueService.ts`, `lib/queues/HandlerContainer.ts`, `lib/queues/MessageSchemaContainer.ts`, and `lib/utils/startupResourcePollingUtils.ts` (mostly `messageAttributes: Record<string, unknown> | undefined` flowing into APIs typed without `| undefined`, plus `messageType: string | undefined`, `logger: CommonLogger | undefined`, etc.). That's a separate refactor that should land on its own — bundling it into a deprecation guard would balloon the PR scope.

2. **The realistic regression is `messageTypeField: '<fieldName>'`, not `: undefined`.** Passing `undefined` is operationally identical to omitting the field — no caller writes that, and if they did, runtime behavior is the same as not passing the option at all. The bug we observed in production was code that hadn't migrated still passing the legacy string value, and that case is now caught.

I improved the diagnostic instead: the option is now typed as a removal-marker string literal `'__REMOVED_USE_messageTypeResolver_INSTEAD__'` so the TS error surfaces the migration hint right at the call site, rather than the generic "not assignable to type 'never'". Pushed as `15ab7f9`.

Happy to land the `exactOptionalPropertyTypes` migration as a follow-up if it's a project goal.

packages/sns/test/SnsPublisherManager.types.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+  'messageSchemas' | 'creationConfig' | 'locatorConfig'
+>
+
+describe('SnsPublisherManager `newPublisherOptions` — legacy `messageTypeField`', () => {
+  it('rejects the removed `messageTypeField` option', () => {
+    const _bad: NewPublisherOptions = {
+      // @ts-expect-error - `messageTypeField` was removed in core 25.x; use `messageTypeResolver` instead
+      messageTypeField: 'type',
+      messageIdField: 'id',
+    }
```
CarlosGamero, line 33: Could we use the expect type of option from vitest?

CarlosGamero (approved):

kibertoad:
Re: CodeRabbit's `exactOptionalPropertyTypes` note — that comment was on the prior `messageTypeField?: never` revision. The current code (15ab7f9) uses a removal-marker string literal `'__REMOVED_USE_messageTypeResolver_INSTEAD__'` with a doc-comment that explicitly calls out the `messageTypeField: undefined` edge case (operationally identical to omitting the field, so the guard catches the realistic regression `messageTypeField: '<name>'`). Enabling `exactOptionalPropertyTypes` repo-wide is out of scope for this fix.

packages/sns/test/SnsPublisherManager.types.spec.ts
kibertoad: Good call — converted the core test (`packages/core/test/types/queueOptionsTypes.types.spec.ts`) to `expectTypeOf` per the convention in `HandlerContainer.types.spec.ts`. Verified it actually fires by intentionally inverting `.not.toMatchTypeOf` and confirming vitest typecheck mode reports the failure.

Decided to **keep `@ts-expect-error`** here in the SNS test, though. The reason: `NewPublisherOptions` is `Omit<SNSPublisherOptions<AnyMessage>, ...>` over a union — and that strips `messageTypeField` from the resolved type's indexed-access keys. I probed with `expectTypeOf<NewPublisherOptions['messageTypeField']>()` and TypeScript reports `Property 'messageTypeField' does not exist on type 'NewPublisherOptions'`. Consequently `expectTypeOf<{ messageTypeField: 'type', ... }>().not.toMatchTypeOf<NewPublisherOptions>()` **passes the wrong way** — structural assignability succeeds because the target effectively doesn't constrain that property anymore.

The thing that actually catches the regression here is TypeScript's **object-literal excess-property check** at the assignment site (`TS2561: Object literal may only specify known properties...`), which `@ts-expect-error` exercises but `expectTypeOf` cannot replicate (it's value-vs-type assignability only, no literal-context excess-property check).

I did enable `typecheck` in `packages/sns/vitest.config.ts` though, so any future `*.types.spec.ts` written using `expectTypeOf` will actually be validated (previously they would have been silent no-ops, which I didn't realize until I tested by intentionally breaking the assertion).

Pushed as `57130d4`.

PR #442, opened 2026-05-18
feat: add zstd message compression codec for SNS and SQS

irfanh94:
## Summary

Adds opt-in **message compression** for the SQS and SNS adapters.

- A `codec` option on **SQS/SNS publishers** compresses outgoing message bodies. Built-in **zstd** uses the Node.js built-in `zlib` module (requires **Node.js >=22.15.0**) — no native dependency, and nothing extra to install. Custom codecs are supported via the `{ name, handler }` form.
- Compressed bodies are wrapped in a self-describing envelope `{ ...routingFields, "__mqtCodec": "zstd", "__mqtData": "<base64>" }`. Identity/routing fields (`id`, `type`, timestamp, deduplication id/options) are copied as plaintext siblings so broker-side filtering (e.g. SNS body-scoped `FilterPolicy`) keeps working on compressed messages.
- **Consumers auto-detect** codec envelopes and decompress transparently before schema validation — built-in codecs need no consumer configuration; custom codecs are registered via `codecs: [...]`, and `disableCodecAutoDetection` opts out. An envelope naming an unregistered codec is surfaced as a **retriable** error (a misconfiguration), not dropped as poison.
- `skipCompressionBelow` (default `512` bytes) skips compression for small messages that would only expand under compression.
- Codec composes with **payload offloading**: the message is compressed **exactly once**; the offload-vs-inline decision is made against the codec envelope wire size; the compress-and-offload path streams through a temp file so large payloads are never fully materialized; `payloadRef.codec` records the algorithm so the consumer decompresses after retrieval with no extra configuration.
- AMQP and Pub/Sub publishers **throw at construction** if `codec` is supplied (unsupported by those adapters).
- `ZstdCodecHandler` caps decompressed size (`maxDecompressedBytes`, default 100 MiB) to guard against decompression bombs.

## Changes

- **`packages/core`** — new `lib/codec/`:
  - `messageCodec.ts` — `MessageCodecEnum`; the `MessageCodec` / `MessageCodecRegistration` / `MessageCodecHandler` / `CodecEnvelope` types; `hasCodecEnvelopeShape` / `isCodecEnvelope`; `KNOWN_CODECS`; `BASE64_RE`.
  - `codecHandler.ts` — `ZstdCodecHandler`, `buildCodecEnvelope`, `resolveCodecHandler`, `getCodecName`.
  - `AbstractQueueService` gains the shared `prepareOutgoingPayload` / `compressAndOffloadPayload` pipeline used by all publishers; `codec` / `skipCompressionBelow` added to `QueuePublisherOptions`; `payloadRef.codec` added to the offloaded-pointer schema. The codec depends only on Node's built-in `zlib`, so it ships **inside core** — there is no separate package to install.
- **`packages/sqs`** — publisher compresses via the shared core pipeline; `AbstractSqsConsumer` auto-detects codec envelopes and decompresses before schema validation, with `codecs[]` / `disableCodecAutoDetection` options; an unregistered codec raises a retriable error.
- **`packages/sns`** — publisher compresses via the shared core pipeline; decompression flows through the shared SQS consumer path.
- **`packages/gcp-pubsub`, `packages/amqp`** — publishers reject the unsupported `codec` option at construction time.
- **build / dev** — workspace devDependencies switched from `"*"` to `"workspace:*"` for consistent local type resolution (pnpm substitutes the concrete version on publish; no impact on npm consumers). Added `bench` scripts/configs for codec micro- and integration benchmarks.

## Test plan

- [x] `pnpm --filter @message-queue-toolkit/core test` — codec unit tests (`messageCodec.spec.ts`, `codecHandler.spec.ts`) and the `AbstractQueueService` offload regression suite.
- [x] `pnpm --filter @message-queue-toolkit/sqs test` — SQS codec integration suites: round-trip + wire-format assertions, message sequences, external compressor, auto-detection, `skipCompressionBelow`, `disableCodecAutoDetection`, custom-codec registration, and codec + payload offloading.
- [x] `pnpm --filter @message-queue-toolkit/sns test` — SNS→SQS codec round-trip, sequence, and auto-detection through the SNS envelope.
- [x] AMQP and Pub/Sub publisher guard tests; existing payload-offloading and consumer suites pass unchanged.

## Summary by CodeRabbit

* **New Features**
  * Add pluggable message compression (built-in zstd + custom codecs) with automatic consumer detection
  * Publisher-side single-pass compression with optional offload of large compressed payloads
  * Configurable compression threshold via skipCompressionBelow
  * New standalone codec package and benchmark suites

* **Documentation**
  * Docs updated with message-compression usage, requirements, and advanced offload behavior

* **Tests**
  * Comprehensive codec integration and unit tests added/expanded

[Review Change Stack](https://app.coderabbit.ai/change-stack/kibertoad/message-queue-toolkit/pull/442?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/core/lib/codec/messageCodec.ts
```diff
@@ -0,0 +1,43 @@
+import { compress, decompress } from '@mongodb-js/zstd'
```
kibertoad: why this implementation specifically?

kibertoad: how heavy is the dependency? I wonder if it belongs in core, which will affect all of the schema packages. ideally core should only store the interface, but not the implementation

irfanh94: @mongodb-js/zstd was chosen for a few reasons:                                                                                                                                                           
    
  1. Native performance — it uses N-API native bindings (prebuilt binaries via prebuild-install), so compression/decompression runs at native speed rather than in JS userland.                            
  2. Battle-tested — it's the same library used in the official MongoDB Node.js driver for wire compression, so it's well-maintained and production-proven at scale.
  3. Simple, focused API — just compress(Buffer): Promise<Buffer> and decompress(Buffer): Promise<Buffer>, nothing extra to configure.                                                                     
  4. Built-in TypeScript types — no separate @types/ package needed.                                                                                                                                       
                                                                                                                                                                                                           
  The main trade-off is the native addon requirement (needs a build step on install, hence the allowBuilds: true entry in pnpm-workspace.yaml). If a pure-JS alternative is preferred for portability,     
  https://github.com/101arrowz/fzstd (WASM-based) would be the closest drop-in, but at the cost of some performance. The codec abstraction in messageCodec.ts is intentionally isolated behind the         
  MessageCodec type and compressMessageBody/decompressMessageBody functions, so swapping the underlying implementation is a one-file change if needed.

irfanh94: Will change it 👍

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -891,6 +893,15 @@ export abstract class AbstractSqsConsumer<
...
       resolveMessageResult.result.body = retrieveOffloadedMessagePayloadResult.result
+    } else if (isCodecEnvelope(resolveMessageResult.result.body)) {
+      try {
+        resolveMessageResult.result.body = await decompressMessageBody(
+          resolveMessageResult.result.body,
+        )
+      } catch (err) {
+        this.handleError(err as Error)
+        return ABORT_EARLY_EITHER
+      }
```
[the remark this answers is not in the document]
irfanh94: Both retry paths use message.Body (the raw SQSMessage string), never resolveMessageResult.result.body:                                                                                                   
                                                                                                                                                                                                           
  - Normal retry — visibility timeout expires; SQS re-delivers the original untouched message from the queue. resolveMessageResult is a local variable that never touches the wire.                        
  - DLQ (failProcessing, line 977-996) — explicitly uses MessageBody: message.Body, the raw SQSMessage.Body string, not the parsed/mutated result.body.                                                    
                                                                                                                                                                                                           
  The overwrite pattern is also already the established precedent in the same method: the offloaded-payload branch at line 895 does exactly the same thing (resolveMessageResult.result.body =             
  retrieveOffloadedMessagePayloadResult.result). The mutation is intentional — it normalises the body for downstream schema validation and parsing in deserializeMessage, without ever feeding that value  
  back to SQS.                                                                                                                                                                                             
                                                                                                                                                                                                         
  No change needed.

packages/core/lib/codec/messageCodec.ts
irfanh94: Changed.

packages/core/lib/codec/codecHandler.ts
```diff
@@ -0,0 +1,36 @@
+import type { CodecEnvelope, MessageCodec, MessageCodecHandler } from '@message-queue-toolkit/core'
+import { compress, decompress } from '@mongodb-js/zstd'
+
+export class ZstdCodecHandler implements MessageCodecHandler {
```
kibertoad, line 31: is there anything SQS-specific to the codec implementation? I would probably introduce a separate package for codecs so that any adapter could use them

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -172,6 +174,7 @@ export abstract class AbstractQueueService<
         }
       : undefined
     this.messageDeduplicationConfig = options.messageDeduplicationConfig
+    this.codec = options.codec
```
kibertoad: documentation needs to be updated to explain entire codec mechanism and available options

packages/core/lib/codec/messageCodec.ts
```diff
@@ -0,0 +1,27 @@
+export const SUPPORTED_CODECS = ['zstd'] as const
+export type MessageCodec = (typeof SUPPORTED_CODECS)[number]
```
CarlosGamero: 🟢 Personally I prefer to use something like:
```
export const MessageCodecEnum = {
  ZSTD: 'zstd'
} as const
export type MessageCodec = ObjectValues<typeof MessageCodecEnum>
```

This allow customers and us to also use `MessageCodecEnum.ZSTD`

packages/core/lib/codec/messageCodec.ts
irfanh94: I removed @mongodb-js/zstd (a native node-gyp addon requiring Python and a C++ toolchain to compile) in favour of zlib.zstdCompress/zstdDecompress built into Node.js 22+. This eliminates the need for build tools in CI and Docker environments, removes 24 transitive packages, and means end users of @message-queue-toolkit/sqs no longer need native build prerequisites just to install the package.

packages/core/lib/codec/messageCodec.ts
irfanh94: Adapted 👍

packages/core/lib/queues/AbstractQueueService.ts
irfanh94: Documented

packages/core/lib/codec/codecHandler.ts
```diff
@@ -0,0 +1,41 @@
+import { promisify } from 'node:util'
+import zlib from 'node:zlib'
+import type { CodecEnvelope, MessageCodec, MessageCodecHandler } from '@message-queue-toolkit/core'
+import { MessageCodecEnum } from '@message-queue-toolkit/core'
+
+const zstdCompress = promisify(zlib.zstdCompress)
+const zstdDecompress = promisify(zlib.zstdDecompress)
+
+export class ZstdCodecHandler implements MessageCodecHandler {
```
CarlosGamero, line 31: Since we’re no longer relying on external libraries for compression, we might consider moving this utility into the core. That would allow us to reuse it across other modules like amp or gcp-pubsup without having to reimplement it each time.

@kibertoad tagging you to get your thoughts on this, especially since you previously suggested moving it out of core 😓

packages/sns/README.md
```diff
@@ -683,6 +684,9 @@ await consumer.start()
   // Optional - Payload Offloading (same as SQS)
   payloadStoreConfig: { /* ... */ },
 
+  // Optional - Compression (Node.js 22+ required)
+  codec: MessageCodecEnum.ZSTD,        // Compress every outgoing message with zstd
```
CarlosGamero, line 688: We only compress the message if it hasn’t been offloaded, right? Let’s clarify this in the readme to avoid confusion.

packages/core/lib/codec/codecHandler.ts
CarlosGamero: Oh, I just saw the comment about a separate package. I agree with that as well—ultimately, we’re both talking about the same thing: enabling compression support across the different technologies we support.

One open question I’m still unsure about: does it really make sense to create a separate package if we’re going to rely on Node internals anyway? If we keep it in core, customers don’t need to install additional dependencies they may not use. From that perspective, having it in core feels sufficient, but I don’t have a strong opinion either way.

pnpm-workspace.yaml
```diff
@@ -6,6 +6,7 @@ packages:
 # pnpm 11 auto-appends new entries here whenever a dep has scripts, prompting
 # an explicit decision per package.
 allowBuilds:
+  '@mongodb-js/zstd': true
```
CarlosGamero: leftover?

packages/sns/README.md
irfanh94: @CarlosGamero this is still in discussion. Once we agree on when compression should be happening I will update the docs 👍

pnpm-workspace.yaml
irfanh94: Yep, removing it in another commit.

packages/codec/lib/codec/codecHandler.ts
```diff
@@ -0,0 +1,41 @@
+import { promisify } from 'node:util'
+import zlib from 'node:zlib'
+import type { CodecEnvelope, MessageCodec, MessageCodecHandler } from '@message-queue-toolkit/core'
+import { MessageCodecEnum } from '@message-queue-toolkit/core'
+
+const zstdCompress = promisify(zlib.zstdCompress)
+const zstdDecompress = promisify(zlib.zstdDecompress)
```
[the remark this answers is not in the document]
CarlosGamero: @irfanh94 this seems important

packages/core/lib/types/queueOptionsTypes.ts
```diff
@@ -139,6 +140,27 @@ export type CommonQueueOptions = {
...
+  /**
+   * Compression codec applied to message bodies.
+   *
+   * - **Publisher**: every outgoing message body is compressed and wrapped in a
+   *   self-describing envelope `{ __codec: 'zstd', __data: '<base64>' }`.
+   * - **Consumer**: when set, the consumer expects compressed messages.
+   *   Even without this option, consumers auto-detect and decompress any message
+   *   that carries a codec envelope, so mixed queues work transparently.
+   *
+   * Uses Node.js built-in `zlib` zstd support — **requires Node.js 22+**.
```
kibertoad: is 22.0.0 OK, or we need to tighten the constraint to the version where it was first released?

packages/core/lib/types/queueOptionsTypes.ts
```diff
@@ -139,6 +140,27 @@ export type CommonQueueOptions = {
   deletionConfig?: DeletionConfig
   payloadStoreConfig?: PayloadStoreConfig
   messageDeduplicationConfig?: MessageDeduplicationConfig
+  /**
+   * Compression codec applied to message bodies.
```
kibertoad: readme needs to be updated as well, and we should recommend using compression in all cases if both sides can understand it

packages/sns/lib/sns/AbstractSnsPublisher.ts
```diff
@@ -211,8 +212,13 @@ export abstract class AbstractSnsPublisher<MessagePayloadType extends object>
     options: SNSMessageOptions,
   ): Promise<void> {
     const attributes = resolveOutgoingMessageAttributes<MessageAttributeValue>(payload)
+    const jsonBody = JSON.stringify(payload)
```
kibertoad: `json-stream-stringify` is a better option here, and compression method ideally should operate on streams, not on full content

packages/core/lib/types/queueOptionsTypes.ts
irfanh94: Readme already updated 👍

packages/sns/README.md
kibertoad: I would suggest compressing it in both cases, as it means faster downloads from the storage as well

packages/sns/lib/sns/AbstractSnsPublisher.ts
irfanh94: @kibertoad Is it safe to use this package: https://www.npmjs.com/package/json-stream-stringify since its been 2 years since last update.

packages/core/lib/types/queueOptionsTypes.ts
irfanh94: It is tightened to 22.15.0

packages/codec/lib/codec/codecHandler.ts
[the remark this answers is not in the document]
irfanh94: @coderabbitai Addressed. Its limited to { "node": ">=22.15.0" }

packages/sns/README.md
irfanh94: Changes pushed:

Upon calling `publish` with codec:
 - compress the data
 - check if the data is over the defined offload limit
 -- if yes: offload compressed binary data (pointer remains uncompressed)
 -- if not: publish compressed message

packages/codec/README.md
```diff
@@ -0,0 +1,84 @@
...
+### Build a codec envelope from already-compressed bytes
+
+When you have pre-compressed bytes (e.g., from `resolveCodecHandler(codec).compress(...)`) and want to produce the envelope string without compressing again:
+
+```typescript
+import { buildCodecEnvelope } from '@message-queue-toolkit/codec'
+import { MessageCodecEnum } from '@message-queue-toolkit/core'
+
+const handler = resolveCodecHandler(MessageCodecEnum.ZSTD)
+const compressed: Buffer = await handler.compress(Buffer.from(JSON.stringify(payload), 'utf8'))
```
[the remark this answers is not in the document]
irfanh94: Addressed

packages/sqs/README.md
```diff
@@ -791,6 +801,54 @@ await publisher.publish({
...
+    }, executionContext)
+  }
+}
+```
+
+#### Notes
+
+- Compression is applied **after** schema validation and **before** the SQS `SendMessage` call.
+- Compressed payloads are still subject to the SQS 256 KB message size limit. For large messages that remain oversized after compression, combine with [Payload Offloading](#payload-offloading).
+- Uses `MessageCodecEnum.ZSTD` (value `'zstd'`). You can use the string literal or the enum — both satisfy the `MessageCodec` type.
```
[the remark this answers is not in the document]
irfanh94: @coderabbitai is this still valid?

packages/sns/lib/sns/AbstractSnsPublisher.ts
kibertoad: should be safe, it's likely just complete, nothing to add. we use it elsewhere in AP too

packages/sqs/lib/sqs/AbstractSqsPublisher.ts
```diff
@@ -199,16 +198,52 @@ export abstract class AbstractSqsPublisher<MessagePayloadType extends object>
...
+   * Returns the payload to send and an optional pre-built body string.
+   * When preBuiltBody is set, it is a ready-to-send codec envelope — sendMessage must use it as-is.
+   */
+  private async prepareOutgoingPayload(message: MessagePayloadType): Promise<{
+    payload: MessagePayloadType | OffloadedPayloadPointerPayload
+    preBuiltBody?: string
+  }> {
+    const codec = this.codec
+
+    if (codec) {
```
kibertoad: **Unconditional compression — small messages get _bigger_.** Every message hits the codec path regardless of size. For tiny payloads (≲200 B) the zstd frame overhead plus the ~33% base64 inflation of the inline envelope means the wire payload grows rather than shrinks. High-throughput small-message systems pay pure CPU + size overhead for no benefit. Consider skipping compression when the raw `Buffer.byteLength` is below a floor (e.g. ~512 B), ideally exposed as a `minSize` codec option.

packages/sqs/lib/sqs/AbstractSqsPublisher.ts
```diff
@@ -199,16 +198,52 @@ export abstract class AbstractSqsPublisher<MessagePayloadType extends object>
...
+      // Compress once up-front, then decide: offload the compressed bytes or send inline.
+      const compressed = await resolveCodecHandler(codec).compress(
+        Buffer.from(JSON.stringify(message), 'utf8'),
```
kibertoad: **Memory regression vs. the streaming offload path.** Pre-PR, large payloads streamed through `JsonStreamStringifySerializer` (stream → temp file → stream upload) — bounded memory for arbitrarily large messages. With codec on, the whole payload is materialised three times right here: `JSON.stringify(message)` (string) → `Buffer.from(..., 'utf8')` (buffer) → `compressed` (buffer). For payloads big enough that streaming offload was the entire point (tens/hundreds of MB), this can OOM the publisher. Consider a true streaming path for the offload-with-codec case: `JsonStreamStringify` → `zlib.createZstdCompress()` → temp file → `store.storePayload`.

packages/sqs/lib/sqs/AbstractSqsPublisher.ts
```diff
@@ -199,16 +198,52 @@ export abstract class AbstractSqsPublisher<MessagePayloadType extends object>
...
+      )
+
+      if (
+        this.payloadStoreConfig &&
+        compressed.byteLength > this.payloadStoreConfig.messageSizeThreshold
```
kibertoad: **Inline path can exceed the SQS 256 KB hard limit.** The inline-vs-offload decision compares `compressed.byteLength` against `messageSizeThreshold`, but the actual wire body is `JSON.stringify({ __codec, __data: base64(compressed) })` — roughly `compressed × 4/3` plus envelope overhead. With `messageSizeThreshold` set at/near SQS's limit, a compressed payload just under the threshold produces a wire body well over 256 KB and SQS rejects the publish at runtime. Compare the **envelope** length (or `Math.ceil(compressed.byteLength * 4 / 3) + ENVELOPE_OVERHEAD`) against the threshold.

Separately — enabling `codec` silently changes `messageSizeThreshold` semantics: without codec it gates on **raw** size, with codec it gates on **compressed** size, so offloading triggers far less often. Worth documenting explicitly, and possibly worth separate thresholds.

packages/sns/lib/sns/AbstractSnsPublisher.ts
```diff
@@ -206,13 +204,51 @@ export abstract class AbstractSnsPublisher<MessagePayloadType extends object>
...
+
+    if (codec) {
+      // Compress once up-front, then decide: offload the compressed bytes or send inline.
+      const compressed = await resolveCodecHandler(codec).compress(
+        Buffer.from(JSON.stringify(message), 'utf8'),
+      )
+
+      if (
+        this.payloadStoreConfig &&
+        compressed.byteLength > this.payloadStoreConfig.messageSizeThreshold
```
kibertoad: **Same 256 KB-ceiling bug as the SQS publisher** (see the comment on `AbstractSqsPublisher.prepareOutgoingPayload`). `compressed.byteLength` is compared, but the inline wire body is the base64 envelope (~33% larger), and SNS has the same 256 KB limit. Fix both call sites together — ideally factor the inline-vs-offload decision into a single shared helper so the two publishers can't drift.

packages/core/lib/codec/messageCodec.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @example
+ * new MyPublisher(deps, { codec: MessageCodecEnum.ZSTD })
+ */
+export const MessageCodecEnum = {
+  /** zstd compression via Node.js built-in `zlib` (requires Node.js >=22.15.0). */
+  ZSTD: 'zstd',
+} as const
+export type MessageCodec = ObjectValues<typeof MessageCodecEnum>
+
+const CODEC_FIELD = '__codec'
```
kibertoad: **`__codec` / `__data` are short, un-namespaced field names.** Two-underscore prefixes collide easily with ORM/framework internals and with user schemas. Combined with the loose `isCodecEnvelope` check below, a collision means a real message is mistaken for a codec envelope. Consider a less collidable namespace, e.g. `__mqtCodec` / `__mqtData`.

kibertoad (commented):
## Review — zstd message compression codec

Reviewed for memory efficiency, performance, bugs, and gaps. Most points are inline; the rest are below. Overall the feature is well-structured (single compression pass, self-describing envelope, auto-detection) and well-tested — but there are a few correctness issues worth resolving before merge.

### P0 — Bugs / data-loss risks
1. **Inline path can exceed the 256 KB SQS/SNS hard limit** — the inline-vs-offload decision compares the *compressed* size, but the wire body is the ~33%-larger base64 envelope. _(inline on both publishers)_
2. **Auto-detection false positives are silently dropped / DLQ'd forever** — `isCodecEnvelope` only checks that two loosely-named fields exist; a legitimate message carrying them is misclassified, fails to decompress, and is lost. _(inline on `messageCodec.ts` and `AbstractSqsConsumer.ts`)_
3. **`payloadRef.codec` is `z.string()`** — accepts unknown codecs that then throw late in the consume path. _(inline)_
4. **Module-load assertion** crashes any importer (incl. type-only) on Node < 22.15. _(inline)_

### P1 — Performance / memory
5. **Compression breaks the streaming-offload memory bound** — the payload is materialised 3× in memory; very large messages that previously streamed to disk can now OOM. _(inline)_
6. **Unconditional compression** makes tiny messages *larger* (frame + base64 overhead). _(inline)_
7. `isCodecEnvelope` rebuilds an array on every consumed message. _(inline)_
8. `buildCodecEnvelope` double-allocates. _(inline)_
9. `streamWithKnownSizeToBuffer` returns a subarray that pins the full allocation. _(inline)_

### P2 — API / extensibility
10. `MessageCodecHandler` is documented as a plug-in point but there's no registry. _(inline)_
11. Enabling `codec` silently changes `messageSizeThreshold` semantics (raw → compressed size). _(noted inline on the threshold check)_
12. `codec` is accepted but silently ignored on Pub/Sub and AMQP publishers. _(inline)_
13. Type-cast bypasses codec validation in the consumer. _(inline)_
14. `payloadRef.size` becomes the compressed size — observability discontinuity. _(inline)_
15. Re-export from `@message-queue-toolkit/sqs` duplicates the public surface. _(inline)_

### P3 — Minor / polish
16. `@types/node ^25` vs `engines >=22.15`. _(inline)_
17. `__codec` / `__data` are short, collision-prone field names. _(inline)_
18. Transient vs. poison failures aren't distinguished on the decompress path. _(inline)_
19. The unrelated `resolveMessage` type-annotation change should be a separate commit. _(inline)_
20. **Missing test:** the README documents that compression can *prevent* offloading (raw size > threshold, compressed size < threshold), but no test asserts S3 is never touched in that case. The wire-format test doesn't combine an inline codec envelope with `payloadStoreConfig` set.
21. The benchmark only prints to console and can't gate regressions in CI. _(inline)_

### Recommended action order
1. Fix #1 (256 KB ceiling) and #2 (false-positive data loss) — these are correctness, not polish.
2. Restore streaming for the offload-with-codec path (#5) and add a `minSize` skip (#6).
3. Tighten validation around `payloadRef.codec` (#3) and the consumer decompress dispatch (#13).
4. Make the codec-handler registry real (#10) before the public API ossifies.
5. Everything else can land as follow-ups.

Happy to help draft patches for any of these.

packages/core/lib/codec/messageCodec.ts
```diff
@@ -0,0 +1,49 @@
...
+export type CodecEnvelope = {
+  [CODEC_FIELD]: MessageCodec
+  [DATA_FIELD]: string
+}
+
+/**
+ * Low-level interface for a compression codec.
+ *
+ * Implement this interface to plug in a custom compression algorithm.
+ * The built-in implementation (`ZstdCodecHandler` in `@message-queue-toolkit/sqs`)
```
kibertoad: Doc nit: this says the built-in handler lives in `@message-queue-toolkit/sqs`, but it's actually in the new `@message-queue-toolkit/codec` package — stale before the PR even merges.

packages/core/lib/codec/messageCodec.ts
```diff
@@ -0,0 +1,49 @@
...
+ * uses Node.js built-in `zlib` zstd support.
+ */
+export interface MessageCodecHandler {
+  compress(data: Buffer): Promise<Buffer>
+  decompress(data: Buffer): Promise<Buffer>
+}
+
+export function isCodecEnvelope(value: unknown): value is CodecEnvelope {
```
kibertoad: **`isCodecEnvelope` is too permissive → false positives.** It returns `true` for _any_ object that merely _contains_ `__codec` (matching the enum) and a string `__data` — extra fields are allowed. Any legitimate message that happens to carry those two fields is misclassified as compressed. Recommend: (a) require a strict 2-key object (`Object.keys(value).length === 2`); (b) provide a consumer opt-out (e.g. `disableCodecAutoDetection`) for schemas that legitimately use these names; (c) validate `__data` is base64-shaped before handing it to the codec. The data-loss consequence is flagged inline on `AbstractSqsConsumer.resolveMaybeOffloadedPayloadMessage`.

packages/core/lib/codec/messageCodec.ts
```diff
@@ -0,0 +1,49 @@
...
+  const record = value as Record<string, unknown>
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    CODEC_FIELD in value &&
+    DATA_FIELD in value &&
+    (Object.values(MessageCodecEnum) as string[]).includes(record[CODEC_FIELD] as string) &&
```
kibertoad: `Object.values(MessageCodecEnum)` allocates a fresh array on every call, and `isCodecEnvelope` runs once per consumed message (hot path). Hoist the allowed values to a module-level `Set` built once.

packages/core/lib/payload-store/offloadedPayloadMessageSchemas.ts
```diff
@@ -11,6 +11,12 @@ export const PAYLOAD_REF_SCHEMA = z.object({
   store: z.string().min(1),
   /** Size of the payload in bytes */
   size: z.number().int().positive(),
+  /**
+   * Codec used to compress the stored payload.
+   * When set, the stored bytes are raw compressed binary (not base64 JSON).
+   * The consumer must decompress using this codec before parsing.
+   */
+  codec: z.string().optional(),
```
kibertoad: **`codec: z.string()` accepts any string.** A pointer with `codec: 'gzip'` (forwarded from a newer producer, or corrupted) passes validation, then throws `Unsupported codec` deep in the consume path and the message is dropped. Tighten to `z.enum([...])` driven by `MessageCodecEnum`, so an unknown codec fails fast as a validation error rather than a late runtime throw.

packages/codec/lib/codec/codecHandler.ts
```diff
@@ -0,0 +1,56 @@
+import { promisify } from 'node:util'
+import zlib from 'node:zlib'
+import type { CodecEnvelope, MessageCodec, MessageCodecHandler } from '@message-queue-toolkit/core'
+import { MessageCodecEnum } from '@message-queue-toolkit/core'
+
+if (typeof zlib.zstdCompress !== 'function' || typeof zlib.zstdDecompress !== 'function') {
```
kibertoad: **Throwing at module-evaluation time is too aggressive.** Any file in the build/import graph that even _type-imports_ from `@message-queue-toolkit/codec` triggers this crash on Node < 22.15 — including type-only consumers and tooling that never actually calls a codec. Make the check lazy: throw inside `ZstdCodecHandler.compress` / `decompress` on first actual use instead of at import.

packages/core/lib/codec/codecHandler.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  decompress(data: Buffer): Promise<Buffer> {
+    return zstdDecompress(data)
+  }
+}
+
+const ZSTD_HANDLER = new ZstdCodecHandler()
+
+export function resolveCodecHandler(codec: MessageCodec): MessageCodecHandler {
+  if (codec === MessageCodecEnum.ZSTD) return ZSTD_HANDLER
```
kibertoad, line 94: **`resolveCodecHandler` is a hardcoded switch, yet `MessageCodecHandler` is documented as a public extension point.** Users implementing the interface have no way to register their handler short of forking. Either add a registry (`registerCodec(name, handler)`) or let the `codec` option accept a `MessageCodecHandler` instance directly alongside the enum string.

packages/codec/lib/codec/codecHandler.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Wraps an already-compressed buffer in a codec envelope string.
+ * Use this when you have pre-compressed bytes and want to avoid compressing twice.
+ */
+export function buildCodecEnvelope(compressed: Buffer, codec: MessageCodec): string {
+  const envelope: CodecEnvelope = {
+    __codec: codec,
+    __data: compressed.toString('base64'),
+  }
+  return JSON.stringify(envelope)
```
kibertoad: `buildCodecEnvelope` does `compressed.toString('base64')` then `JSON.stringify` — two extra full-payload string allocations per publish, so the inline path peaks at ≈4× the compressed size. Unavoidable for inline transport, but for payloads near the threshold you could skip the `JSON.stringify` pass and emit the envelope directly: `'{"__codec":"' + codec + '","__data":"' + b64 + '"}'`.

packages/codec/lib/codec/codecHandler.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export async function decompressMessageBody(envelope: CodecEnvelope): Promise<unknown> {
+  const handler = resolveCodecHandler(envelope.__codec)
+  const compressed = Buffer.from(envelope.__data, 'base64')
```
kibertoad: `Buffer.from(str, 'base64')` is lenient — it silently drops non-base64 characters instead of failing. A non-envelope message that slipped past `isCodecEnvelope` reaches here and produces garbage bytes (then a confusing zstd error) rather than a clean "not a valid envelope" error. Validate `__data` is well-formed base64 before decoding.

packages/core/lib/utils/streamUtils.ts
```diff
@@ -14,5 +14,10 @@ export async function streamWithKnownSizeToString(stream: Readable, size: number
     offset += chunkBuffer.length
   }
 
-  return buffer.toString('utf8', 0, offset)
+  return buffer.subarray(0, offset)
```
kibertoad: `buffer.subarray(0, offset)` returns a view that still pins the full `Buffer.alloc(size)` allocation. If the stream ever yields fewer bytes than `size`, the returned `Buffer` retains the larger backing store. Minor, and only matters on a size mismatch — but copying (`Buffer.from(buffer.subarray(0, offset))`) or slicing only when `offset !== size` would be safer.

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -915,19 +917,32 @@ export abstract class AbstractSqsConsumer<
...
-    if (!resolveMessageResult.result || !resolveMessageResult.result.body) {
+    if (!resolveMessageResult.result?.body) {
       return ABORT_EARLY_EITHER
     }
 
     if (hasOffloadedPayload(resolveMessageResult.result)) {
       const retrieveOffloadedMessagePayloadResult = await this.retrieveOffloadedMessagePayload(
         resolveMessageResult.result.body,
+        (codec, data) => {
+          const handler = resolveCodecHandler(codec as Parameters<typeof resolveCodecHandler>[0])
```
kibertoad: `codec as Parameters<typeof resolveCodecHandler>[0]` casts away validation. `codec` here comes from `payloadRef.codec`, which is currently `z.string()` (see the schema comment) — an unknown value passes the cast and throws inside `resolveCodecHandler`. Validate explicitly (e.g. an `isKnownCodec(codec)` guard) and surface a structured error rather than casting.

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -915,19 +917,32 @@ export abstract class AbstractSqsConsumer<
...
+          return handler.decompress(data)
+        },
       )
       if (retrieveOffloadedMessagePayloadResult.error) {
         this.handleError(retrieveOffloadedMessagePayloadResult.error)
         return ABORT_EARLY_EITHER
       }
       resolveMessageResult.result.body = retrieveOffloadedMessagePayloadResult.result
+    } else if (isCodecEnvelope(resolveMessageResult.result.body)) {
```
kibertoad: **Auto-detection false positive = silent data loss.** If `isCodecEnvelope` misfires on a legitimate message (see the `isCodecEnvelope` comment), `decompressMessageBody` throws → `handleError` + `ABORT_EARLY_EITHER` → `failProcessing`, which routes to the DLQ if one is configured or **silently ACKs and drops the message otherwise**. A real, valid message is lost permanently. Tightening `isCodecEnvelope` and adding an opt-out is the mitigation; also consider distinguishing "looks like an envelope but failed to decompress" from a genuine poison message so retry/DLQ policy can differ.

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -730,14 +707,82 @@ export abstract class AbstractQueueService<
...
+    compressed: Buffer,
+    codec: MessageCodec,
+  ): Promise<OffloadedPayloadPointerPayload> {
+    if (!this.payloadStoreConfig) {
+      throw new Error('Payload store is not configured')
+    }
+
+    const { store, storeName } = this.resolveOutgoingStore()
+    const payloadId = await store.storePayload({
+      value: Readable.from(compressed),
```
kibertoad: `Readable.from(compressed)` wraps an already fully-materialised buffer — the streaming interface here is cosmetic, the whole compressed payload is in memory. Combined with the publisher-side materialisation (see the `AbstractSqsPublisher` comment), the offload-with-codec path no longer has the bounded-memory property the plain offload path has. A true streaming compress (`JsonStreamStringify` → `createZstdCompress` → temp file) would restore it.

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -730,14 +707,82 @@ export abstract class AbstractQueueService<
...
+      size: compressed.byteLength,
+    })
+
+    return this.buildPointer(message, payloadId, storeName, compressed.byteLength, codec)
```
kibertoad: `payloadRef.size` (and the legacy `offloadedPayloadSize` / `OFFLOADED_PAYLOAD_SIZE_ATTRIBUTE`) is now the **compressed** byte length when codec is active. Any metrics/dashboards tracking offloaded payload size will show a discontinuity at the codec/no-codec boundary. Consider also recording the uncompressed size (e.g. `payloadRef.uncompressedSize`) or at least documenting the change.

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -787,6 +832,24 @@ export abstract class AbstractQueueService<
...
+      try {
+        const compressedBuffer = await streamWithKnownSizeToBuffer(
+          serializedOffloadedPayloadReadable,
+          payloadSize,
+        )
+        const decompressed = await decompress(codec, compressedBuffer)
+        return { result: JSON.parse(decompressed.toString('utf8')) }
+      } catch (e) {
+        return {
+          error: new Error(`Failed to decompress offloaded payload with codec "${codec}"`, {
```
kibertoad, line 1148: All failures inside this block — truncated S3 stream, transient read error, genuinely corrupt bytes — collapse into one generic error and (on the SQS path) get treated as a non-retriable poison message. A transient store-read failure should ideally be retriable. Consider distinguishing decompression/parse failures (poison) from retrieval failures (retry).

packages/gcp-pubsub/lib/pubsub/AbstractPubSubPublisher.ts
```diff
@@ -69,11 +69,12 @@ export abstract class AbstractPubSubPublisher<MessagePayloadType extends object>
       const parsedMessage = messageSchemaResult.result.parse(message)
 
       message = this.updateInternalProperties(message)
-      const maybeOffloadedPayloadMessage = await this.offloadMessagePayloadIfNeeded(message, () => {
-        // Calculate message size for PubSub
-        const messageData = Buffer.from(JSON.stringify(message))
-        return messageData.length
-      })
+      const maybeOffloadedPayloadMessage =
```
kibertoad, line 78: **`codec` is silently ignored for Pub/Sub (and AMQP).** `codec` lives on `CommonQueueOptions`, so every publisher's constructor accepts it, but only the SQS/SNS publishers act on it — this Pub/Sub publisher only got the `offloadPayload` return-type update. A user who sets `codec` on a Pub/Sub or AMQP publisher gets no compression and no error. Either implement it here too, or throw at construction when `codec` is set on a publisher that doesn't support it.

packages/sqs/lib/index.ts
```diff
@@ -1,3 +1,9 @@
+export {
```
kibertoad: Re-exporting codec utilities from `@message-queue-toolkit/sqs` creates a second public import path for symbols that already live in `@message-queue-toolkit/codec`. Two import paths for the same API complicate future deprecation and invite version skew. Prefer having consumers import from `@message-queue-toolkit/codec` directly and drop the re-export.

packages/codec/package.json
```diff
@@ -0,0 +1,58 @@
...
+    },
+    "peerDependencies": {
+        "@message-queue-toolkit/core": ">=25.0.0"
+    },
+    "devDependencies": {
+        "@biomejs/biome": "^2.3.8",
+        "@lokalise/biome-config": "^3.1.0",
+        "@lokalise/tsconfig": "^3.0.0",
+        "@message-queue-toolkit/core": "workspace:*",
+        "@types/node": "^25.0.2",
```
kibertoad: Minor: `@types/node` is `^25` while `engines.node` is `>=22.15.0`. Usually harmless, but pinning the dev types to a range matching the lowest supported runtime avoids accidentally compiling against APIs newer than Node 22.

packages/sqs/bench/codec.bench.ts
```diff
@@ -0,0 +1,198 @@
...
+import { SqsPermissionConsumer } from '../test/consumers/SqsPermissionConsumer.ts'
+import type { PERMISSIONS_ADD_MESSAGE_TYPE } from '../test/consumers/userConsumerSchemas.ts'
+import { SqsPermissionPublisher } from '../test/publishers/SqsPermissionPublisher.ts'
+import type { TestAwsResourceAdmin } from '../test/utils/testAdmin.ts'
+import type { Dependencies } from '../test/utils/testContext.ts'
+import { registerDependencies } from '../test/utils/testContext.ts'
+
+// ─── Configuration ────────────────────────────────────────────────────────────
+
+const N = 50
```
kibertoad, line 29: Useful smoke benchmark, but it only prints to console — it can't catch a regression in CI. Consider saving a baseline (msg/s, overhead %) and asserting against it, or wiring it into `vitest bench` with thresholds. Also, `N = 50` against LocalStack is dominated by network round-trips, so the compression CPU cost is buried in the noise — a separate micro-benchmark of `compress`/`decompress` alone would show the actual codec overhead.

packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
```diff
@@ -201,7 +207,9 @@ export abstract class AbstractSnsSqsConsumer<
     await this.startConsumers()
   }
 
-  protected override resolveMessage(message: SQSMessage) {
+  protected override resolveMessage(
```
kibertoad, line 214: This `resolveMessage` return-type tightening is unrelated to the codec feature. Not wrong, but it makes the PR harder to review and to revert independently — ideally a separate commit/PR.

irfanh94: You're right that it's unrelated — it slipped in as a drive-by while the file was open. Happy to keep it here since it's a no-op (TypeScript already inferred the same type), but let me know if you'd prefer I revert it and land it separately.

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -730,14 +715,132 @@ export abstract class AbstractQueueService<
...
+          pointer: this.buildPointer(message, payloadId, storeName, compressedSize, codecName),
+        }
+      }
+
+      // Compressed payload fits inline — return the buffer; caller wraps it in a codec envelope.
+      return { compressedBuffer: fs.readFileSync(tmpPath) }
+    } finally {
+      try {
+        fs.unlinkSync(tmpPath)
+      } catch {
```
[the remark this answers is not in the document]
irfanh94: Addressed

packages/core/lib/codec/codecHandler.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Wraps an already-compressed buffer in a codec envelope string.
+ * Use this when you have pre-compressed bytes and want to avoid compressing twice.
+ *
+ * Uses string concatenation instead of JSON.stringify to avoid allocating an
+ * intermediate object — the base64 string and the envelope string are the only
+ * two allocations on the inline path.
+ */
+export function buildCodecEnvelope(compressed: Buffer, codecName: string): string {
+  return '{"__mqtCodec":"' + codecName + '","__mqtData":"' + compressed.toString('base64') + '"}'
+}
```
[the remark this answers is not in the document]
irfanh94: Addressed in the latest commit

kibertoad (commented):
## Review: zstd compression codec

Thorough, well-tested feature with a clean commit history. Inline comments below are prioritized **[P1]** high / **[P2]** medium / **[P3]** low. Two non-inline items:

- **PR description is stale.** It still shows the `{ "__codec", "__data" }` envelope (renamed to `__mqtCodec`/`__mqtData`), `@mongodb-js/zstd` (replaced by built-in `zlib`), `packages/core/lib/codec/messageCodec.ts` as the impl home (moved to `packages/codec`), and the sqs barrel re-export (later removed). Worth refreshing before merge.
- **No UPGRADING.md / changelog entry** for the new `@message-queue-toolkit/codec` package or the new core exports.

The two most important items are the SNS consumer option leak and the stale consumer docs — both stem from the `codec` → `codecs` consumer-API refactor not being fully propagated.

packages/sns/test/consumers/SnsSqsPermissionConsumer.ts
```diff
@@ -41,6 +41,7 @@ type SnsSqsPermissionConsumerOptions = Pick<
   | 'maxRetryDuration'
   | 'payloadStoreConfig'
   | 'concurrentConsumersAmount'
+  | 'codec'
```
kibertoad: **[P1] SNS consumers silently re-expose `codec`/`skipCompressionBelow`, defeating the consumer-API redesign.**

The SQS consumer deliberately does `Omit<QueueConsumerOptions<…>, 'codec' | 'skipCompressionBelow'>` and replaces them with a `codecs` array. But `SNSSQSConsumerOptions = SQSConsumerOptions & SNSOptions & {…}`, and `SNSOptions = QueueOptions<…>` re-includes the full `CommonQueueOptions` — so `codec` and `skipCompressionBelow` come back onto SNS consumer options. That's the only reason `Pick<…, 'codec'>` compiles here.

Concrete harm: passing a **custom** codec as `codec: { name, handler }` to an SNS consumer type-checks and sets `this.codec`, but the consumer decompresses via `codecRegistry` (built from `options.codecs`). The handler is never registered → custom-codec messages fail `isCodecEnvelope` scoping and silently route to the DLQ.

Also: this test and `SnsSqsPermissionConsumer.codec.spec.ts` pass `codec: 'zstd'` to the consumer — that only "works" because built-in zstd is auto-registered regardless; the option itself is dead.

Fix: apply the same `Omit` to the SNS consumer options (or strip them in `AbstractSnsSqsConsumer`), plumb `codecs` through this helper, and add an SNS custom-codec test.

packages/core/README.md
```diff
@@ -641,6 +641,54 @@ class MyPayloadStore implements PayloadStore {
...
+
+class MyLz4Handler implements MessageCodecHandler {
+  async compress(data: Buffer): Promise<Buffer> { /* ... */ }
+  async decompress(data: Buffer): Promise<Buffer> { /* ... */ }
+  createCompressStream(): Transform { /* return a Transform stream */ }
+}
+
+const codec = { name: 'lz4', handler: new MyLz4Handler() }
+new MyPublisher(deps, { codec })
+new MyConsumer(deps, { codec }) // same registration required on the consumer
```
kibertoad: **[P1] Stale consumer docs.** Consumer options were changed from `codec` to a `codecs` array. `new MyConsumer(deps, { codec })` should be `new MyConsumer(deps, { codecs: [codec] })`. (Built-in zstd needs no consumer option at all — only custom codecs need `codecs`.)

packages/codec/README.md
```diff
@@ -0,0 +1,112 @@
...
+Register the handler on the publisher and consumer using the `{ name, handler }` object form:
+
+```typescript
+const codec = { name: 'lz4', handler: new MyLz4Handler() }
+
+// Publisher — wraps each outgoing message in { __mqtCodec: 'lz4', __mqtData: '<base64>' }
+new MyPublisher(deps, { codec })
+
+// Consumer — only auto-detects envelopes whose __mqtCodec matches 'lz4'
+new MyConsumer(deps, { codec })
```
kibertoad: **[P1] Stale consumer docs.** Consumer registration is the `codecs` array now: `new MyConsumer(deps, { codecs: [codec] })`, not `{ codec }`.

packages/sqs/README.md
```diff
@@ -788,16 +799,80 @@ await publisher.publish({
...
+
+```typescript
+import { MessageCodecEnum } from '@message-queue-toolkit/core'
+
+class MyConsumer extends AbstractSqsConsumer<SupportedMessages, ExecutionContext> {
+  constructor(deps: SQSConsumerDependencies) {
+    super(deps, {
+      // Optional: explicitly declare that messages are compressed.
+      // Without this, consumers still auto-detect and decompress codec envelopes.
+      codec: MessageCodecEnum.ZSTD,
```
kibertoad: **[P1] Stale / non-compiling consumer example.** SQS consumer options **omit `codec`** (replaced by `codecs`), so this snippet would not compile if copied. For built-in zstd a consumer needs **no** option — auto-detection handles it. The "explicitly declare that messages are compressed" comment describes an option that no longer exists. Use `codecs: [...]` only for custom codecs.

packages/sqs/README.md
```diff
@@ -538,6 +544,11 @@ When using `locatorConfig`, you connect to an existing queue without creating it
     payloadStore: s3Store,
   },
 
+  // Optional - Compression (Node.js >=22.15.0 required)
+  // Auto-detection is always active: consumers decompress codec envelopes
+  // even without this option set.
+  codec: MessageCodecEnum.ZSTD,
```
kibertoad: **[P1] Stale consumer config reference.** `codec` is not a valid SQS consumer option (omitted in favor of `codecs`). Remove it here, or show `codecs: [...]` for the custom-codec case.

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -730,14 +715,132 @@ export abstract class AbstractQueueService<
...
+    codecName: string,
+  ): Promise<
+    | { pointer: OffloadedPayloadPointerPayload; compressedBuffer?: never }
+    | { compressedBuffer: Buffer; pointer?: never }
+  > {
+    if (!this.payloadStoreConfig) {
+      throw new Error('Payload store is not configured')
+    }
+
+    const tmpPath = `${os.tmpdir()}/${randomUUID()}`
```
kibertoad: **[P2] `codec` + `payloadStoreConfig` writes a temp file to disk for _every_ message, regardless of size.**

The streaming temp-file path exists to avoid 3× buffering of _large_ payloads, but it runs unconditionally whenever a store is configured. For the common case (small messages, store present only as a safety net), this is a disk write + read per publish — a throughput/IO regression vs. the old `offloadMessagePayloadIfNeeded`, which size-checked first and returned early.

Consider compressing in-memory for payloads below a bound (e.g. the protocol max) and reserving the temp-file path for genuinely large payloads.

Minor: build the path with `node:path.join` rather than string `/` concatenation.

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -730,14 +715,132 @@ export abstract class AbstractQueueService<
...
+    try {
+      // Streaming pipeline: serializer output → codec transform → temp file.
+      // No full-payload buffer is materialised; each codec supplies its own Transform.
+      await pipeline(
+        typeof serialized.value === 'string' ? Readable.from(serialized.value) : serialized.value,
+        handler.createCompressStream(),
+        fs.createWriteStream(tmpPath),
+      )
+
+      const compressedSize = fs.statSync(tmpPath).size
```
kibertoad: **[P2] Synchronous fs in an async hot path.** `fs.statSync` (here), `fs.readFileSync` (L819), and `fs.unlinkSync` (L822) block the event loop — `readFileSync` on up to ~`messageSizeThreshold` bytes per publish. Use `fs.promises` (`stat`/`readFile`/`unlink`). Notable in a PR framed around performance.

packages/core/lib/queues/AbstractQueueService.ts
```diff
@@ -730,14 +715,132 @@ export abstract class AbstractQueueService<
...
+
+      // Compare the envelope wire size (not raw compressed bytes) against the threshold.
+      // buildCodecEnvelope produces {"__mqtCodec":"<codecName>","__mqtData":"<base64>"}.
+      // Base64 expands by ⌈N/3⌉×4; the fixed JSON framing adds 32 chars + codec name length.
+      const envelopeSize = Math.ceil((compressedSize * 4) / 3) + 32 + codecName.length
```
kibertoad: **[P3] Base64 size estimate is slightly off and contradicts its own commit message.** Actual base64 length is `Math.ceil(N / 3) * 4`, not `Math.ceil(N * 4 / 3)` — the commit message itself states "⌈N/3⌉×4". The current form underestimates by up to 2 bytes, which only matters right at the protocol limit, but it's easy to make exact.

packages/sqs/lib/sqs/AbstractSqsPublisher.ts
```diff
@@ -123,9 +124,7 @@ export abstract class AbstractSqsPublisher<MessagePayloadType extends object>
       // (offloaded payload won't have user fields needed for messageGroupIdField)
       const resolvedOptions = this.resolveFifoOptions(message, options)
 
-      const maybeOffloadedPayloadMessage = await this.offloadMessagePayloadIfNeeded(message, () =>
-        calculateOutgoingMessageSize(message),
-      )
+      const { payload, preBuiltBody } = await this.prepareOutgoingPayload(message)
```
kibertoad, line 139: **[P2] Compression + S3 upload happen before the dedup short-circuit.** `prepareOutgoingPayload()` compresses (and on the codec+offload path writes a temp file and uploads to S3) _before_ the `isDeduplicationEnabledForMessage` check below. A deduplicated message still pays full compression CPU, and the codec+offload path leaves an **orphan S3 object**. `parsedMessage` is already available — move the dedup check ahead of `prepareOutgoingPayload`. (Orphan S3 on plain offload is pre-existing; the wasted compression is new.)

packages/sns/lib/sns/AbstractSnsPublisher.ts
```diff
@@ -206,13 +204,69 @@ export abstract class AbstractSnsPublisher<MessagePayloadType extends object>
...
 
+  /**
+   * Compresses (when codec is set) or offloads (when store is configured) the message.
+   * Returns the payload to send and an optional pre-built body string.
+   * When preBuiltBody is set, it is a ready-to-send codec envelope — sendMessage must use it as-is.
+   *
+   * When both codec and payloadStoreConfig are set, uses a streaming pipeline
+   * (JSON → zstd → temp file → store) to avoid materialising the full payload in memory.
+   */
+  private async prepareOutgoingPayload(message: MessagePayloadType): Promise<{
```
kibertoad: **[P2] `prepareOutgoingPayload` is duplicated verbatim** between this file and `AbstractSqsPublisher` (~45 identical lines). Since `compressAndOffloadPayload` already lives in core, hoist `prepareOutgoingPayload` there too so the two copies can't drift.

packages/codec/package.json
```diff
@@ -0,0 +1,58 @@
...
+        "clean": "rimraf dist",
+        "lint": "biome check . && tsc",
+        "lint:fix": "biome check --write .",
+        "prepublishOnly": "pnpm run lint && pnpm run build"
+    },
+    "engines": {
+        "node": ">=22.15.0"
+    },
+    "peerDependencies": {
+        "@message-queue-toolkit/core": ">=25.0.0"
```
kibertoad: **[P2] Peer-dependency floor is too loose.** This package imports `MessageCodecEnum`, `MessageCodecHandler`, `MessageCodecRegistration`, `CodecEnvelope` from core — all **new** in this PR. `>=25.0.0` allows installing against core 25.0.0–25.5.0, which lack those exports → runtime import failure. Bump the floor to the core version that ships the codec types.

packages/sqs/package.json
```diff
@@ -36,6 +37,7 @@
     },
     "peerDependencies": {
         "@aws-sdk/client-sqs": "^3.1034.0",
+        "@message-queue-toolkit/codec": ">=1.0.0",
         "@message-queue-toolkit/core": ">=25.0.0",
```
kibertoad: **[P2] Peer-dependency floor too loose.** sqs now imports `KNOWN_CODECS`, `isCodecEnvelope`, `MessageCodecHandler`, `MessageCodecRegistration` from core — new in this PR. `@message-queue-toolkit/core: ">=25.0.0"` allows an older in-range core that lacks them. Bump to the core version shipping these APIs.

packages/sns/package.json
```diff
@@ -38,6 +38,7 @@
         "@aws-sdk/client-sns": "^3.632.0",
         "@aws-sdk/client-sqs": "^3.1034.0",
         "@aws-sdk/client-sts": "^3.632.0",
+        "@message-queue-toolkit/codec": ">=1.0.0",
         "@message-queue-toolkit/core": ">=24.0.0",
```
kibertoad: **[P2] Peer-dependency floor too loose.** `@message-queue-toolkit/core: ">=24.0.0"` is well below the core version that introduces the codec APIs this feature relies on (transitively via sqs). Bump the floor accordingly.

packages/core/README.md
```diff
@@ -641,6 +641,54 @@ class MyPayloadStore implements PayloadStore {
...
+See the [`@message-queue-toolkit/codec` README](../codec/README.md) for a full custom codec example.
+
+Compressed messages are wrapped in a self-describing envelope `{ __mqtCodec: '<name>', __mqtData: '<base64>' }`. Consumers configured with a matching `codec` registration decompress transparently. Consumers without a matching registration ignore the envelope — `codec` does not need to be set when using the built-in zstd and you are happy with auto-detection.
+
+#### Interaction with codec (compression)
+
+When both `codec` and `payloadStoreConfig` are set on a publisher, compression and offloading work together with a single compression pass:
+
+1. The message is compressed **once** at publish time.
+2. The **compressed** size is compared against `messageSizeThreshold`.
```
kibertoad: **[P2] Doc contradicts the code.** `compressAndOffloadPayload` compares the estimated **codec envelope wire size** (base64 + JSON framing) against `messageSizeThreshold`, and `payloadStoreTypes.ts` JSDoc says so. This line (and the equivalent prose in `packages/sqs/README.md` and `packages/codec/README.md`) says the **compressed size** is compared. Align the READMEs with the code/JSDoc.

packages/sqs/test/consumers/SqsPermissionConsumer.codec.spec.ts
```diff
@@ -0,0 +1,625 @@
...
+    await wirePublisher.publish(message)
+
+    const result = await wireConsumer.handlerSpy.waitForMessageWithId(message.id, 'consumed')
+    expect(result.message).toMatchObject(message)
+
+    await wirePublisher.close()
+    await wireConsumer.close(true)
+  })
+
+  it('consumer with disableCodecAutoDetection passes codec-shaped message through as plain JSON', async () => {
```
kibertoad, line 451: **[P2] This test doesn't actually exercise `disableCodecAutoDetection`.** As the test's own comment admits, the publisher sends plain JSON here (skipCompressionBelow prevents compression), so the body is never a real codec envelope. It only proves the flag doesn't break the plain-JSON path. The actual behavior — a genuine `{__mqtCodec,__mqtData}` envelope passing through **un-decompressed** when the flag is set — is untested. Add a case that publishes a real envelope to a queue and asserts the consumer treats it as a raw object.

packages/core/lib/types/queueOptionsTypes.ts
```diff
@@ -139,6 +140,60 @@ export type CommonQueueOptions = {
...
+   * Set to `0` to compress every message regardless of size.
+   *
+   * @example
+   * // Compress only messages ≥ 1 KB
+   * new MyPublisher(deps, { codec: MessageCodecEnum.ZSTD, skipCompressionBelow: 1024 })
+   *
+   * // Always compress (disable the floor)
+   * new MyPublisher(deps, { codec: MessageCodecEnum.ZSTD, skipCompressionBelow: 0 })
+   */
+  skipCompressionBelow?: number
```
kibertoad: **[P2] Incomplete docs: `skipCompressionBelow` is silently ignored when a payload store is configured.** `compressAndOffloadPayload` intentionally skips this floor (documented only in an internal code comment). A user setting `skipCompressionBelow: 99999` alongside `payloadStoreConfig` will be surprised that small messages are still compressed. State this caveat in the user-facing JSDoc here and in the READMEs.

packages/sqs/lib/sqs/AbstractSqsConsumer.ts
```diff
@@ -915,19 +948,40 @@ export abstract class AbstractSqsConsumer<
...
       }
       resolveMessageResult.result.body = retrieveOffloadedMessagePayloadResult.result
+    } else if (
+      !this.disableCodecAutoDetection &&
+      isCodecEnvelope(resolveMessageResult.result.body, this.codecKnownNames)
+    ) {
+      try {
+        const envelope = resolveMessageResult.result.body
+        const handler = this.codecRegistry.get(envelope.__mqtCodec)
+        if (!handler) throw new Error(`Unknown codec: ${envelope.__mqtCodec}`)
```
kibertoad: **[P3] Dead check.** `isCodecEnvelope(body, this.codecKnownNames)` on the `else if` above already guarantees `__mqtCodec ∈ codecKnownNames`, and `codecKnownNames` is exactly `codecRegistry.keys()` — so `codecRegistry.get(envelope.__mqtCodec)` can't be `undefined` here. The `if (!handler) throw` is unreachable. Harmless, but removable.

packages/amqp/lib/AbstractAmqpPublisher.ts
```diff
@@ -51,6 +51,14 @@ export abstract class AbstractAmqpPublisher<
     dependencies: AMQPDependencies,
     options: AMQPPublisherOptions<MessagePayloadType, CreationConfig, LocatorConfig>,
   ) {
+    // `codec` lives on CommonQueueOptions (added after this package's core peer dep was frozen),
+    // so it is not present in the local type.  The cast guards JS callers and future versions.
+    if ((options as { codec?: unknown }).codec) {
```
kibertoad: **[P3] Inconsistency.** `sqs`, `sns`, and `gcp-pubsub` migrated their `@message-queue-toolkit/core` devDependency to `workspace:*`, but `amqp` did not — which is why this guard needs the `as { codec?: unknown }` cast while the PubSub guard uses a clean `options.codec`. Migrating amqp's core devDep too would let this cast go away and keep the packages consistent.

packages/codec/lib/codec/codecHandler.ts
```diff
@@ -0,0 +1,97 @@
...
+} from '@message-queue-toolkit/core'
+import { MessageCodecEnum } from '@message-queue-toolkit/core'
+
+/**
+ * Validates that a string is properly-padded base64 before passing it to Buffer.from.
+ * Buffer.from(str, 'base64') silently ignores non-base64 characters, so without this
+ * check a malformed __mqtData field produces garbage bytes and a confusing codec error
+ * instead of a clear "invalid envelope" message.
+ */
+const BASE64_RE =
```
kibertoad: **[P3] Duplicated regex.** This `BASE64_RE` is identical to the one in `packages/core/lib/codec/messageCodec.ts`. Export one and reuse it to avoid the two definitions drifting.

packages/core/lib/codec/codecHandler.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export async function decompressMessageBody(envelope: CodecEnvelope): Promise<unknown> {
```
kibertoad: **[P3] `decompressMessageBody` can't handle custom codecs.** `resolveCodecHandler` throws `Unsupported codec` for any non-`zstd` string, so this exported util only works for built-in codecs. The consumer path is unaffected (it uses its registry), but a user calling this public util on a custom-codec envelope gets a confusing error. Either document the built-in-only limitation or accept an optional handler/registry.

PR #445, opened 2026-05-25
chore: SNS support for SQS 26

CarlosGamero:
## Release Notes

* **Breaking Changes**
  * SNS service now uses grouped resource getters instead of individual fields; attempting to access before initialization will throw an error.
  * Added new `areResourcesReady` getter for checking resource initialization status.
  * Requires SQS peer dependency `>=26.0.0`.

* **Documentation**
  * Added upgrade guide documenting breaking changes in SNS/SQS service resource handling and new initialization patterns.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kibertoad/message-queue-toolkit/pull/445?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
+++ b/packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
@@ -165,7 +165,7 @@ export abstract class AbstractSnsSqsConsumer<
               if (this.startRequested) {
                 this.logger.info({
                   message: 'Resources now ready, starting consumers',
-                  queueName: this.creationConfig?.queue.QueueName ?? this.locatorConfig?.queueName,
+                  queueName: result.queueName,
                   topicArn: result.topicArn,
                 })
                 return this.startConsumers()
@@ -224,17 +224,26 @@ export abstract class AbstractSnsSqsConsumer<
       // Resources not ready yet (non-blocking polling mode), mark that start was requested.
       // Consumers will be started automatically when onResourcesReady callback fires.
       this.startRequested = true
-      this.logger.info({
-        message:
-          'Start requested but resources not ready yet, will start when resources become available',
-        queueName: this.creationConfig?.queue.QueueName ?? this.locatorConfig?.queueName,
-        topicArn: this.locatorConfig?.topicArn,
-      })
+      this.logger.info(
+        this.buildConfiguredResourceLogContext(),
+        'Start requested but resources not ready yet, will start when resources become available',
+      )
       return
     }
 
     // Resources are ready, start consumers immediately
     await this.startConsumers()
+  }
+
+  private buildConfiguredResourceLogContext(): Record<string, string> {
+    const ctx: Record<string, string> = {}
+    const queueName = this.creationConfig?.queue.QueueName ?? this.locatorConfig?.queueName
+    const topicArn = this.locatorConfig?.topicArn
+    const topicName = this.creationConfig?.topic?.Name ?? this.locatorConfig?.topicName
+    if (queueName) ctx.queueName = queueName
+    if (topicArn) ctx.topicArn = topicArn
+    else if (topicName) ctx.topicName = topicName
+    return ctx
   }
 
   public override async close(abort?: boolean): Promise<void> {
--- a/packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
+++ b/packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
@@ -237,6 +237,13 @@ export abstract class AbstractSnsSqsConsumer<
     await this.startConsumers()
   }
 
+  public override async close(abort?: boolean): Promise<void> {
+    this._subscription = undefined
+    this.resourcesReady = false
+    this.startRequested = false
+    await super.close(abort)
+  }
+
   protected override resolveMessage(
     message: SQSMessage,
   ): Either<MessageInvalidFormatError | MessageValidationError, ResolvedMessage> {
--- a/packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
+++ b/packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
@@ -210,6 +210,7 @@ export abstract class AbstractSnsSqsConsumer<
 
   protected setSubscriptionResource(resource: SubscriptionResource): void {
     this._subscription = resource
+    this.isInitted = true
   }
 
   /**
--- a/packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
+++ b/packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
@@ -133,9 +133,12 @@ export abstract class AbstractSnsSqsConsumer<
         onResourcesReady: (result) => {
           // Update values that were empty when resourcesReady was false
           this.topicArn = result.topicArn
-          this.queueUrl = result.queueUrl
           this.subscriptionArn = result.subscriptionArn
-          this.queueName = result.queueName
+          this.setQueueResource({
+            name: result.queueName,
+            url: result.queueUrl,
+            arn: result.queueArn,
+          })
           this.resourcesReady = true
 
           // Initialize DLQ now that resources are ready (this is mutually exclusive
@@ -152,7 +155,7 @@ export abstract class AbstractSnsSqsConsumer<
               if (this.startRequested) {
                 this.logger.info({
                   message: 'Resources now ready, starting consumers',
-                  queueName: this.queueName,
+                  queueName: this.creationConfig?.queue.QueueName ?? this.locatorConfig?.queueName,
                   topicArn: this.topicArn,
                 })
                 return this.startConsumers()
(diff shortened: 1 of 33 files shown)


packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
```diff
@@ -79,10 +84,13 @@ export abstract class AbstractSnsSqsConsumer<
...
-  protected topicArn: string
-  // @ts-expect-error
-  protected subscriptionArn: string
+  /**
+   * Resolved topic + subscription handle. Populated together by `initSnsSqs`
+   * (either synchronously or via the non-blocking `onResourcesReady`
+   * callback). Subclasses read via the {@link subscription} getter so the
+   * ARNs are impossible to access in an "uninitialised" state.
+   */
+  private _subscription?: SubscriptionResource
```
kibertoad, line 93: **P0 — `close()` leaks this field, defeating the stated invariant.**

The class inherits `AbstractSqsConsumer.close()` → `AbstractSqsService.close()`, which only resets `_queue` and `isInitted`. `_subscription` (added here) is never cleared, nor are `resourcesReady` or `startRequested`. So immediately after `close()`:

- `consumer.subscription.topicArn` still returns the stale ARN — the docstring above says the ARN should be *"impossible to access in an 'uninitialised' state"*, but it is.
- `resourcesReady === true` can mislead a re-init code path.
- `startRequested` may incorrectly carry across init cycles.

Suggested fix — add an override:

```ts
public override async close(abort?: boolean): Promise<void> {
  this._subscription = undefined
  this.resourcesReady = false
  this.startRequested = false
  await super.close(abort)
}
```

kibertoad (commented):
Drive-by review of the SQS-26 mirroring work. Two P0 correctness issues called out inline (close() leak, `queueArn!` assertion), plus a few P1/P2 follow-ups. A combined UPGRADING.md migration note for the SQS-26 + SNS protected-field rename would also be worth adding before merge.

packages/sns/lib/utils/snsInitter.ts
```diff
@@ -345,19 +342,35 @@ export async function initSnsSqs(
...
     let queueAvailable = false
+    let queueArn: string | undefined
 
     const notifyIfBothReady = () => {
       if (nonBlocking && topicAvailable && queueAvailable) {
         extraParams?.onResourcesReady?.({
           topicArn: subscriptionTopicArn,
           queueUrl,
+          // biome-ignore lint/style/noNonNullAssertion: queueAvailable is checked above
+          queueArn: queueArn!,
```
kibertoad: **P0 — `queueArn!` is unsafe.**

`markQueueAvailable` (just below) sets `queueAvailable = true` whenever `getQueueAttributes` doesn't *throw*, but `attrs.result?.attributes?.QueueArn` can be `undefined` (the `Attributes` map only contains requested names, and SDK responses with an empty map have been seen in the wild). When that happens, `onResourcesReady` fires with `queueArn: undefined`, `setQueueResource` happily stores it, and any later `this.queue.arn` access returns `undefined` — silently.

Mirror the explicit check used on the synchronous path at the bottom of `initSnsSqs`:

```ts
if (!queueArn) {
  extraParams?.onResourcesError?.(
    new Error(`Could not resolve QueueArn for queue at ${queueUrl}`),
    { isFinal: true },
  )
  return
}
queueAvailable = true
```

…then the non-null assertion here goes away.

packages/sns/lib/utils/snsInitter.ts
```diff
@@ -345,19 +342,35 @@ export async function initSnsSqs(
...
           // subscriptionArn is guaranteed to be defined here because we're in the branch where locatorConfig.subscriptionArn exists
           subscriptionArn: locatorConfig.subscriptionArn as string,
           queueName,
         })
       }
     }
 
+    const markQueueAvailable = async () => {
+      try {
+        const attrs = await getQueueAttributes(sqsClient, queueUrl, ['QueueArn'])
```
kibertoad: **P2 — extra `GetQueueAttributes` round-trip per init.**

`pollForQueue`'s `checkFn` already calls `getQueueAttributes(queueUrl)` with default `['All']` (which includes `QueueArn`) but throws the result away because `pollForQueue` is typed `Promise<unknown | undefined>`. Same applies to the synchronous `getQueueAttributes(..., ['QueueArn'])` at L482.

Threading the result through (change `pollForQueue` return type to `Promise<QueueAttributesResult | undefined>`) lets you reuse it and eliminates both extra round-trips. Modest gain per startup but it compounds with the new background polling path.

packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
```diff
@@ -152,8 +165,8 @@ export abstract class AbstractSnsSqsConsumer<
               if (this.startRequested) {
                 this.logger.info({
                   message: 'Resources now ready, starting consumers',
```
kibertoad, line 167: **P1 — log loses context for locator-only setups.**

When the consumer was wired with `locatorConfig.topicName` (not `topicArn`), the resolved ARN isn't reachable here yet (`_subscription` is set in the same callback, but the log is built before `this.subscription` is safe to read — and even after, you can't read it for the `startRequested` branch at L229).

Same issue at L229: `topicArn: this.locatorConfig?.topicArn` will be `undefined` for any locator using `topicName`, or for a locator with only `subscriptionArn`.

Prefer reading from `result.topicArn` here (already in scope), and at L229 either resolve and cache the topic ARN once via `buildTopicArn(...)` for diagnostics, or omit the field when unavailable rather than emitting `undefined`.

packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
```diff
@@ -168,22 +181,35 @@ export abstract class AbstractSnsSqsConsumer<
...
+
+    await this.initDeadLetterQueue()
+  }
+
+  protected get subscription(): Readonly<SubscriptionResource> {
+    if (!this._subscription) throw new Error('Subscription is not started yet')
+    return this._subscription
+  }
+
+  protected setSubscriptionResource(resource: SubscriptionResource): void {
```
kibertoad, line 211: **P3 — asymmetry: `setQueueResource` and `setTopicResource` flip `isInitted`, this one doesn't.**

It's correct today because `setSubscriptionResource` is always paired with `setQueueResource` in the flow, but the next maintainer who reaches for this from a custom subclass will silently leave `isInitted = false`. Either:

- add a one-line comment explaining the asymmetry, or
- make it set `isInitted = true` for symmetry (cheap idempotent set).

packages/sns/test/consumers/SnsSqsPermissionConsumer.ts
```diff
@@ -179,12 +179,26 @@ export class SnsSqsPermissionConsumer extends AbstractSnsSqsConsumer<
...
   get subscriptionProps() {
-    return {
-      topicArn: this.topicArn,
-      queueUrl: this.queueUrl,
-      queueName: this.queueName,
-      subscriptionArn: this.subscriptionArn,
-      deadLetterQueueUrl: this.deadLetterQueueUrl,
+    // topicArn / subscriptionArn / _queue are populated together by init;
+    // if the queue resource isn't resolved yet, none of the others are either.
+    try {
```
kibertoad: **P1 — this try/catch pattern is duplicated across 5 consumer files and conflates two failure modes.**

Same block appears in `SnsSqsPermissionConsumer.ts`, `SnsSqsPermissionConsumerFifo.ts`, `SnsSqsEntityConsumer.ts`, `CreateLocateConfigMixConsumer.ts`, and the polling spec's `TestStartupResourcePollingConsumer`. Two suggestions:

1. Lift it into `AbstractSnsSqsConsumer` (e.g., `protected getResourceHandles(): { … } | undefined` or an `isResourcesReady` getter) so test consumers don't each catch errors.
2. Inside the helper, prefer probing private state (`if (!this._subscription || !this._queue) return undefined`) over try/catch — that way unexpected getter exceptions still bubble up instead of being silently swallowed into an all-`undefined` shape that looks like "not initialised yet".

packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
CarlosGamero: Thanks for catching this — totally right. Added a `close()` override in 4d9d523 that resets `_subscription`, `resourcesReady` and `startRequested` before delegating to `super.close(abort)`.

packages/sns/lib/utils/snsInitter.ts
CarlosGamero: Thanks, good spot. Fixed in 4d9d523: `markQueueAvailable` now reports via `onResourcesError` (final) when the QueueArn fetch returns an empty Attributes map, instead of silently flipping the queue to "ready". Took the chance to drop the separate `queueAvailable` flag and use `queueArn` truthiness as the ready signal, so the `!` assertion in `notifyIfBothReady` goes away too.

packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
CarlosGamero: Thanks. Fixed in 2dd1123:

- L167 callback log now reads `queueName` from `result` too (same source of truth as `result.topicArn` already was).
- L229 `start()` log uses a new `buildConfiguredResourceLogContext()` helper that emits only the identifiers we can derive from config — `queueName`, plus `topicArn` or `topicName` as fallback — and omits the field entirely when unknown (so locator-only-with-`subscriptionArn` no longer logs `topicArn: undefined`).

Skipped the `buildTopicArn` cache path for now — felt heavier than the diagnostic warrants.

packages/sns/test/consumers/SnsSqsPermissionConsumer.ts
CarlosGamero: Thanks. Done in 3efb46e:

- Added `protected get areResourcesReady` on `AbstractSnsSqsConsumer` (exposes the existing private `resourcesReady` flag).
- All five test consumers' `subscriptionProps` getters now short-circuit on `!this.areResourcesReady` and read the throwing `queue`/`subscription` getters directly when ready. No more try/catch, so unexpected exceptions propagate as you suggested instead of being flattened into an all-`undefined` shape that looks like "not initialised yet".

packages/sns/lib/utils/snsInitter.ts
CarlosGamero: Thanks. Done in c4e60ab:

- `QueueAttributesResult` is now exported from `@message-queue-toolkit/sqs`.
- `pollForQueue` is typed `Promise<QueueAttributesResult | undefined>` and threads the attrs through `onResourceAvailable(attrs)`.
- `markQueueAvailable` is now a sync helper that reads `QueueArn` from the attrs it receives — no separate fetch.
- The non-polling branch (`Promise.all`) captures `queueCheckResult.result` into a shared `queueAttributes` variable.
- The L482 final-return fetch is gone — it reads `QueueArn` from the captured `queueAttributes` instead.

Net: 2 fewer round-trips per init in the locator-only polling path, 1 fewer in the non-polling path.

packages/sns/lib/sns/AbstractSnsSqsConsumer.ts
CarlosGamero: Thanks. Done in 9f2328a — `setSubscriptionResource` now flips `isInitted = true` for symmetry with `setQueueResource` / `setTopicResource`. Idempotent at the existing call sites (always paired with `setQueueResource`).

kibertoad (approved):

PR #484, opened 2026-06-18
Bump @platformatic/kafka version

kjamrog:
## Summary

Bumps `@platformatic/kafka` to `^2.3.1` (and Node engine to `>= 22.22.0`).

## Topics deletion fix

`@platformatic/kafka` 2.3.x has a regression in `Admin#deleteTopics`: on the error path it bypasses the deduplication callback, leaving a stale entry in the per-instance dedup cache. Reusing the same `Admin` afterwards causes subsequent `deleteTopics` calls for the same topics to hang forever.

This broke the `init - close` tests, which share a singleton `kafkaAdmin` and delete the same topics in `beforeEach`: the first delete (topics don't exist yet) errors and pollutes the cache, so every later delete times out. Tests passed individually but failed when run together.

Workaround: a shared `deleteExistingTopics` test helper lists topics first and only deletes the ones that exist, avoiding the broken error path. The upstream fix is being pushed separately.

### DIFF
--- a/packages/kafka/package.json
+++ b/packages/kafka/package.json
@@ -2,7 +2,7 @@
     "name": "@message-queue-toolkit/kafka",
     "version": "0.11.4",
     "engines": {
-        "node": ">= 22.14.0"
+        "node": ">= 22.22.0"
     },
     "private": false,
     "license": "MIT",
@@ -53,7 +53,7 @@
     "dependencies": {
         "@lokalise/node-core": "^14.8.1",
         "@lokalise/universal-ts-utils": "^4.10.0",
-        "@platformatic/kafka": "^2.2.3"
+        "@platformatic/kafka": "^2.3.1"
     },
     "peerDependencies": {
         "@message-queue-toolkit/core": ">=26.1.1",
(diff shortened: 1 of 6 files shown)


packages/kafka/package.json
```diff
@@ -2,7 +2,7 @@
     "name": "@message-queue-toolkit/kafka",
     "version": "0.11.4",
     "engines": {
-        "node": ">= 22.14.0"
+        "node": ">= 22.22.0"
```
kjamrog, line 5: To be compatible with `@platformatic/kafka` requirement

packages/kafka/test/utils/testContext.ts
```diff
@@ -55,6 +55,23 @@ export const getKafkaConfig = (): KafkaConfig => ({
   clientId: randomUUID(),
 })
 
+/**
+ * Deletes only the topics that currently exist.
+ *
+ * Workaround for a regression in @platformatic/kafka >= 2.3.0 where Admin#deleteTopics
```
kjamrog, line 61: It should be a simple fix for @platformatic/kafka, I will create PR for it. Since we only use topics deletion in tests and the workaround is clean (checking existing topics before deletion makes sense), it should be fine to proceed with the version upgrade

kibertoad (approved):

packages/kafka/test/utils/testContext.ts
kjamrog: PR for the library: https://github.com/platformatic/kafka/pull/323

PR #560, opened 2026-09-01
feat: require zod 4.5 and precompile registered schemas

kibertoad:
## What

Raises the required `zod` version to 4.5 and uses what landed there: ahead-of-time schema compilation.

- **Peer floor bumped to `>=4.5.0 <5.0.0`** in `core`, `amqp`, `sqs`, `sns`, `gcp-pubsub` and `kafka`. Dev dependencies moved to `^4.5.4`. `schemas` keeps its `>=3.25.67` floor: it imports nothing from 4.5, and raising it would break standalone installs for no gain.
- **Every schema passed to the toolkit is precompiled automatically**, at the point where it is registered:
  - `MessageSchemaContainer`, both the publisher `messageSchemas` side and the consumer side, so SQS / SNS / AMQP / Pub-Sub / Kafka all get it, and both the schema map and the definition map
  - `MessageHandlerConfig` and `MessageHandlerConfigBuilder.addConfig`
  - `KafkaHandlerConfig`
  - `EventRegistry`, for the definitions `DomainEventEmitter.emit` parses with
  - the toolkit's own per-message schemas: `OFFLOADED_PAYLOAD_POINTER_PAYLOAD_SCHEMA`, which parses on every message once a payload store is configured, and `MESSAGE_DEDUPLICATION_OPTIONS_SCHEMA`
- **Compilation is memoized per schema.** A typical wiring registers the same event definition three or four times over: `EventRegistry` compiles it, `AbstractPublisherManager` hands the same `consumerSchema` to a publisher's `MessageSchemaContainer`, and a consumer handler registers it again. A `WeakMap` keyed on the schema the caller owns turns every registration after the first into a lookup.

`precompileSchema` and `precompileEventDefinition` are exported from `@message-queue-toolkit/core`.

## Why

On a representative event schema, compiled parsing runs ~2.5x faster than the interpreted parser (200k parses: 38ms -> 14ms), for ~0.2ms of one-off compilation per schema. Consumers parse every message they receive and publishers parse every message they send, so the cost is paid once at registration and repaid on every message.

Compilation lives at registration rather than at schema-definition time on purpose: only schemas that actually reach a publisher, a consumer or a registry get compiled, so definitions an app never wires up cost nothing at import.

## Behavior notes

`compile()` returns a clone, so the schema the toolkit parses with is no longer the object that was passed in. `MessageSchemaContainer.resolveSchema()`, `MessageHandlerConfig.schema`, `KafkaHandlerConfig.schema` and `EventRegistry.getEventDefinitionByTypeName()` return the precompiled counterpart. `EventRegistry.supportedEvents` still holds the array exactly as it was given. Because the memo makes the clone stable, `expect(config.schema).toBe(precompileSchema(MY_SCHEMA))` is the way to assert on it.

Parsing behavior is unchanged, including errors: a schema zod refuses to compile (an async refinement, for instance) is handed back untouched, keeps using the regular parser, and the refusal is memoized so repeat registrations do not re-run codegen that will refuse again.

Precompilation never mutates a schema the caller owns.

`UPGRADING.md` and the README cover all of this.

## Release label

This is labeled `major`. Tightening a peer range is a breaking change: a consumer on zod 4.4 taking what looks like a safe upgrade would get an unsatisfiable peer range and a failed install. `packages/kafka` peers on `@message-queue-toolkit/core` `>=27.0.0` accordingly.

## Unrelated CI fix

`gcp-pubsub` could not typecheck on a clean install, and had not been able to since the `@google-cloud/storage` 8.0.1 bump in #554. It pulled `@message-queue-toolkit/gcs-payload-store` from the registry, whose peer range resolved storage 7.22.0, while its own tests picked up the hoisted 8.0.1, leaving two copies of `Storage` that are mutually unassignable. Only the `gcs-payload-store` jobs ran on #554, so nothing caught it there.

Fixed by pointing `gcp-pubsub` at the workspace `gcs-payload-store` (as `sqs` and `sns` already do for `s3-payload-store`), giving it an explicit `@google-cloud/storage` dev dependency so resolution does not depend on hoist order, and widening the `gcs-payload-store` peer range to cover the 8.x it is already tested against. One copy of storage remains in the tree, and the lockfile drops ~300 lines.

## Testing

`turbo run build` and `turbo run lint` pass for all 12 packages.

| suite | result |
| --- | --- |
| core | 233 passed, no type errors |
| schemas | 22 passed, no type errors |
| kafka (real broker) | 88 passed, no type errors |
| gcp-pubsub (emulators) | 98 passed |
| gcs-payload-store (emulator) | 11 passed |
| amqp (rabbitmq) | 63 passed |
| metrics | 29 passed |
| outbox-core | 5 passed |
| s3-payload-store | 10 passed |

New coverage: `precompileUtils.spec.ts` covers compilation, the memo, the non-compilable fallback and its memo, non-mutation, and automatic precompilation at each registration point including the container's definition map. `KafkaHandlerConfig.spec.ts` is new.

Not run locally: `sqs`, `sns` and `redis-message-deduplication-store`. Ports 4566 and 6379 were held by unrelated containers. Those suites were green on CI for the previous push.

## Summary by CodeRabbit

* **New Features**
  * Schemas and event definitions are now automatically precompiled when registered or configured.
  * Precompilation is memoized, improving consistency and avoiding repeated compilation work.
  * Added the exported `precompileSchema` helper for independently precompiling schemas.
* **Breaking Changes**
  * Supported integrations now require Zod `^4.5.0`; Zod 3 is no longer supported.
  * Schema-related inspection types and helper APIs are no longer publicly exported.
* **Documentation**
  * Updated schema precompilation and upgrade guidance, including changes to returned schema instances.


packages/amqp/package.json
```diff
@@ -37,7 +37,7 @@
         "@message-queue-toolkit/core": ">=25.0.0",
         "@message-queue-toolkit/schemas": ">=7.0.0",
         "amqplib": "^2.0.1",
-        "zod": ">=3.25.76 <5.0.0"
+        "zod": ">=4.5.0 <5.0.0"
```
CatchMe2: We could achieve the same range with `^4.5.0` (it would be consistent with amqplib range definition)
We could follow this approach for other packages changes in this PR as well.

CatchMe2 (approved):

packages/amqp/package.json
CarlosGamero: Yeah, it is basically the same but shorter definition

CarlosGamero (approved):

packages/amqp/package.json
kibertoad: Agreed — `^4.5.0` is exactly `>=4.5.0 <5.0.0` for a non-zero major, so the accepted set is unchanged, and it matches how every other third-party peer here is written (`amqplib` `^2.0.1`, `@google-cloud/pubsub` `^6.0.1`, `@aws-sdk/*` `^3.x`); open-ended `>=` stays for the internal `@message-queue-toolkit/*` peers.

Applied across all six packages this PR touches — `core`, `amqp`, `sqs`, `sns`, `gcp-pubsub`, `kafka` — plus the restatement in `UPGRADING.md`. No lockfile change, since workspace peer ranges aren't recorded there.

Two ranges I left alone, as a caret can't express either: `schemas` keeps `zod` `>=3.25.67` (spans majors 3 and 4) and `gcs-payload-store` keeps `@google-cloud/storage` `>=7.17.0 <9.0.0` (spans 7.x and 8.x).

I don't have push access to this branch, so the change is in #564, targeting `feat/precompile-schemas-zod-4.5` — merging it folds the change into this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #578, opened 2026-09-07
feat(schemas): allow object-or-union payloads in event definitions

CarlosGamero:
## What

`enrichMessageSchemaWithBase` / `enrichMessageSchemaWithBaseStrict` constrained the payload schema to a `ZodObject`, so an event payload could not be modelled as a **union of object variants** (e.g. a single-item / multi-item shape).

This relaxes the constraint to a new `EventPayloadSchema = z.ZodType<Record<string, unknown>>`:

- a plain object schema — still accepted
- a union of object schemas — now accepted
- a bare scalar payload (`z.string()`, `z.number()`, …) — still rejected

`CommonEventDefinition`'s sentinel payload is widened the same way so union-payload events still satisfy it. Defining the type via the *output* (`Record<string, unknown>`) rather than `ZodObject | ZodUnion` avoids `.extend` distributing over a type-level union and collapsing it back to a plain object.

## Why

Enables events that carry either one item or many (union of the current single shape and an array-of-items shape) without abandoning the object-based payload guarantee.

## Tests

`baseMessageSchemas.spec.ts`:
- union payload accepted, with a real `consumerSchema.parse` of a multi-item message
- scalar payload rejected (`@ts-expect-error`)

`tsc --noEmit` clean, full schemas suite green (24/24).

## AI Assistance Tracking

We're running a metric to understand where AI assists our engineering work. Please select exactly one of the options below:

Mark "Yes" if AI helped in any part of this work, for example: generating code, refactoring, debugging support, explaining something, reviewing an idea, or suggesting an approach.

- [x] **Yes, AI assisted with this PR**
- [ ] **No, AI did not assist with this PR**

## Summary by CodeRabbit

* **New Features**
  * Event and message schema helpers now support unions of object payload variants and object-producing transforms.
  * Unsupported scalar payloads and schemas with unrestricted outputs are rejected during validation.

* **Documentation**
  * Added upgrade guidance for the expanded payload support and updated function type-argument requirements.

* **Tests**
  * Added coverage for union payloads, transformed payloads, validation rules, and runtime parsing.

### DIFF
--- a/packages/schemas/lib/events/eventTypes.ts
+++ b/packages/schemas/lib/events/eventTypes.ts
@@ -1,21 +1,11 @@
 import { z } from 'zod/v4'
 
 import { MetadataObjectSchema } from '../messages/baseMessageSchemas.ts'
-import { CONSUMER_BASE_EVENT_SCHEMA, PUBLISHER_BASE_EVENT_SCHEMA } from './baseEventSchemas.ts'
-
-/**
- * Schema allowed as an event payload: anything whose input and output are objects, which covers
- * a plain object schema and a union of object variants (e.g. a single-item / multi-item shape).
- * It stays object-based (a bare string/number payload is not allowed) while accepting unions.
- * Defined via the input/output types (not `ZodObject | ZodUnion`) so `.extend` does not distribute
- * over a type-level union and collapse it back to a plain object.
- *
- * Both the output AND input type parameters are pinned: leaving `Input` at its default `unknown`
- * would widen `z.input<publisherSchema>['payload']` to `unknown`, dropping the object-payload
- * guarantee for code written generically over `CommonEventDefinition`, and would let
- * scalar-input transforms (e.g. `z.preprocess`) through.
- */
-export type EventPayloadSchema = z.ZodType<Record<string, unknown>, Record<string, unknown>>
+import {
+  CONSUMER_BASE_EVENT_SCHEMA,
+  type EventPayloadSchema,
+  PUBLISHER_BASE_EVENT_SCHEMA,
+} from './baseEventSchemas.ts'
 
 export type EventTypeNames<EventDefinition extends CommonEventDefinition> =
   CommonEventDefinitionConsumerSchemaType<EventDefinition>['type']
--- a/packages/schemas/lib/events/eventTypes.ts
+++ b/packages/schemas/lib/events/eventTypes.ts
@@ -4,13 +4,18 @@ import { MetadataObjectSchema } from '../messages/baseMessageSchemas.ts'
 import { CONSUMER_BASE_EVENT_SCHEMA, PUBLISHER_BASE_EVENT_SCHEMA } from './baseEventSchemas.ts'
 
 /**
- * Schema allowed as an event payload: anything whose output is an object, which covers a
- * plain object schema and a union of object variants (e.g. a single-item / multi-item shape).
+ * Schema allowed as an event payload: anything whose input and output are objects, which covers
+ * a plain object schema and a union of object variants (e.g. a single-item / multi-item shape).
  * It stays object-based (a bare string/number payload is not allowed) while accepting unions.
- * Defined via the output type (not `ZodObject | ZodUnion`) so `.extend` does not distribute
+ * Defined via the input/output types (not `ZodObject | ZodUnion`) so `.extend` does not distribute
  * over a type-level union and collapse it back to a plain object.
+ *
+ * Both the output AND input type parameters are pinned: leaving `Input` at its default `unknown`
+ * would widen `z.input<publisherSchema>['payload']` to `unknown`, dropping the object-payload
+ * guarantee for code written generically over `CommonEventDefinition`, and would let
+ * scalar-input transforms (e.g. `z.preprocess`) through.
  */
-export type EventPayloadSchema = z.ZodType<Record<string, unknown>>
+export type EventPayloadSchema = z.ZodType<Record<string, unknown>, Record<string, unknown>>
 
 export type EventTypeNames<EventDefinition extends CommonEventDefinition> =
   CommonEventDefinitionConsumerSchemaType<EventDefinition>['type']
--- a/packages/schemas/lib/events/eventTypes.ts
+++ b/packages/schemas/lib/events/eventTypes.ts
@@ -3,6 +3,15 @@ import { z } from 'zod/v4'
 import { MetadataObjectSchema } from '../messages/baseMessageSchemas.ts'
 import { CONSUMER_BASE_EVENT_SCHEMA, PUBLISHER_BASE_EVENT_SCHEMA } from './baseEventSchemas.ts'
 
+/**
+ * Schema allowed as an event payload: anything whose output is an object, which covers a
+ * plain object schema and a union of object variants (e.g. a single-item / multi-item shape).
+ * It stays object-based (a bare string/number payload is not allowed) while accepting unions.
+ * Defined via the output type (not `ZodObject | ZodUnion`) so `.extend` does not distribute
+ * over a type-level union and collapse it back to a plain object.
+ */
+export type EventPayloadSchema = z.ZodType<Record<string, unknown>>
+
 export type EventTypeNames<EventDefinition extends CommonEventDefinition> =
   CommonEventDefinitionConsumerSchemaType<EventDefinition>['type']
 
@@ -12,11 +21,11 @@ export function isCommonEventDefinition(entity: unknown): entity is CommonEventD
 
 const consumerSchema = CONSUMER_BASE_EVENT_SCHEMA.extend({
   metadata: MetadataObjectSchema,
-  payload: z.looseObject({}),
+  payload: z.looseObject({}) as EventPayloadSchema,
 })
 
 const publisherSchema = PUBLISHER_BASE_EVENT_SCHEMA.extend({
-  payload: z.looseObject({}),
+  payload: z.looseObject({}) as EventPayloadSchema,
 })
 
 export type CommonEventDefinition = {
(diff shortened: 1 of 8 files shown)


packages/schemas/lib/events/eventTypes.ts
```diff
@@ -12,11 +21,11 @@ export function isCommonEventDefinition(entity: unknown): entity is CommonEventD
 
 const consumerSchema = CONSUMER_BASE_EVENT_SCHEMA.extend({
   metadata: MetadataObjectSchema,
-  payload: z.looseObject({}),
+  payload: z.looseObject({}) as EventPayloadSchema,
```
kibertoad, line 19: `z.ZodType<Output>` leaves its second parameter (`Input`) at the default `unknown`, so `EventPayloadSchema` constrains only the output side. Casting the sentinel payload to it makes `z.input<typeof publisherSchema>['payload']` resolve to `unknown`, where on `main` it was `{ [k: string]: unknown }`.

Verified A/B with this probe dropped into `packages/schemas/lib`:

```ts
declare function publishGeneric<E extends CommonEventDefinition>(
  def: E,
  data: CommonEventDefinitionPublisherSchemaType<E>,
): void
declare const someDef: CommonEventDefinition
publishGeneric(someDef, {
  type: 'anything',
  payload: 'this-is-not-an-object',
  metadata: { schemaVersion: '1', producedBy: 'a', originatedFrom: 'a', correlationId: 'c' },
})
```

On `main` this is `error TS2322: Type 'string' is not assignable to type '{ [x: string]: unknown; }'`. On this branch it compiles clean. So any helper, or an `AbstractPublisherManager` / `DomainEventEmitter` wrapper, written generically over `CommonEventDefinition` rather than over a concrete event union silently loses the object-payload guarantee. The consumer side is unaffected: `z.output` payload is still `Record<string, unknown>`.

Pinning the input parameter fixes it:

```ts
export type EventPayloadSchema = z.ZodType<Record<string, unknown>, Record<string, unknown>>
```

I checked that this two-parameter form still accepts the new union payload, plain objects, `z.looseObject`, `z.strictObject`, `z.discriminatedUnion`, `z.intersection`, `.brand()`, `.partial()`, `.pick()` and `.catchall()`, and still rejects `z.string()` and `z.array(...)`.

packages/schemas/lib/events/eventTypes.ts
```diff
@@ -3,6 +3,15 @@ import { z } from 'zod/v4'
...
 import { CONSUMER_BASE_EVENT_SCHEMA, PUBLISHER_BASE_EVENT_SCHEMA } from './baseEventSchemas.ts'
 
+/**
+ * Schema allowed as an event payload: anything whose output is an object, which covers a
+ * plain object schema and a union of object variants (e.g. a single-item / multi-item shape).
+ * It stays object-based (a bare string/number payload is not allowed) while accepting unions.
+ * Defined via the output type (not `ZodObject | ZodUnion`) so `.extend` does not distribute
+ * over a type-level union and collapse it back to a plain object.
+ */
+export type EventPayloadSchema = z.ZodType<Record<string, unknown>>
```
kibertoad: Because the constraint bounds only the output type, two things that were compile errors before now pass:

```ts
// payload resolves to `any` for this event, so every handler and publisher loses payload checking
enrichMessageSchemaWithBase('any', z.any())

// the input side is unconstrained, so a scalar-input payload is accepted
const pre = enrichMessageSchemaWithBase(
  'pre',
  z.preprocess((v) => ({ a: String(v) }), z.object({ a: z.string() })),
)
// z.input<typeof pre.publisherSchema>['payload'] is `number`, so `payload: 42` typechecks
```

`z.string().transform((s) => ({ a: s }))` behaves the same way. Both compile on this branch and both were rejected on `main`.

The description's "a bare scalar payload is still rejected" holds only for the literal `z.string()` / `z.number()` form the new test covers. Pinning the input parameter (see the comment on line 24) closes the transform/preprocess hole. `z.any()` would need an explicit exclusion if that case matters.

packages/schemas/lib/messages/baseMessageSchemas.ts
```diff
@@ -79,19 +78,19 @@ export type SchemaMetadata = {
   description: string
 }
 
-export function enrichMessageSchemaWithBaseStrict<
-  T extends ZodObject<Y>,
-  Y extends ZodRawShape,
-  Z extends string,
->(type: Z, payloadSchema: T, schemaMetadata: SchemaMetadata): ReturnType<T, Y, Z> {
+export function enrichMessageSchemaWithBaseStrict<T extends EventPayloadSchema, Z extends string>(
```
kibertoad, line 83: Dropping `Y extends ZodRawShape` changes the explicit type-argument arity of three exported functions from 3 to 2: `enrichMessageSchemaWithBaseStrict` here, `enrichMessageSchemaWithBase` (line 89) and `getMessageType` (line 113).

Verified: `enrichMessageSchemaWithBase<typeof s, { a: z.ZodString }, 'x'>('x', s)` no longer compiles on this branch, while `enrichMessageSchemaWithBase<typeof s, 'x'>('x', s)` does. Inference-only call sites are fine (I typechecked `packages/{schemas,core,sns,sqs,amqp,outbox-core,kafka,metrics}` clean against this branch), but downstream code passing explicit type args breaks.

The PR carries the `patch` label and `publish.yml` derives the version bump from it, so this would ship as a patch with no `UPGRADING.md` entry. A `minor` bump plus a short migration note looks warranted.

packages/schemas/lib/messages/baseMessageSchemas.spec.ts
```diff
@@ -3,16 +3,49 @@ import { z } from 'zod/v4'
...
+    producedBy: ['USER_SERVICE'],
+  },
 } as const satisfies Record<string, SnsAwareEventDefinition>
 
 describe('base message schemas', () => {
   it('message satisfies SnsAwareEventDefinition', () => {
     expectTypeOf(myEvents.myEvent).toExtend<SnsAwareEventDefinition>()
   })
+
+  it('rejects a non-object payload schema', () => {
```
kibertoad: This calls `enrichMessageSchemaWithBase('user.updated', z.string())` under `@ts-expect-error` with no `expect`, so at runtime it asserts nothing while reading as a runtime test.

It also pins only the one case the new constraint still catches. It would not have caught `z.any()` as a payload, a scalar-input `.transform()` / `z.preprocess()` payload, or the widening of `z.input<publisherSchema>['payload']` to `unknown`, all of which this branch admits. `expectTypeOf` cases for those would make the guard match the claim in the description.

kibertoad (commented):
Reviewed the diff for correctness. Notes inline, plus one on an unchanged file that cannot be anchored.

The main issue is that `EventPayloadSchema` constrains only the output side of the schema, which widens the publisher payload input type to `unknown` and lets `z.any()` and scalar-input transforms through. Everything else I probed (unions, `.brand()`, `.partial()`, `.pick()`, `.catchall()`, `strictObject`, `looseObject`, `discriminatedUnion`, `intersection`, arrays and bare scalars) behaves as intended, and the repo typechecks clean end to end.

---

**`packages/schemas/lib/events/baseEventSchemas.ts:108`** (not in this diff, so no inline anchor)

`enrichEventSchemaWithBase` is the sibling of the helper being relaxed and is re-exported from `lib/index.ts`, but it and its local `ReturnType<T, Y, Z>` (line 92) still require `T extends ZodObject<Y>`.

After this change a union payload compiles through `enrichMessageSchemaWithBase` and fails to compile through `enrichEventSchemaWithBase`, with nothing to hint that the other entry point accepts it. Worth relaxing both, or noting the asymmetry.

packages/schemas/lib/events/eventTypes.ts
CarlosGamero: Fixed — `EventPayloadSchema` now pins both parameters: `z.ZodType<Record<string, unknown>, Record<string, unknown>>`, so `z.input<publisherSchema>["payload"]` stays object-typed. (67eb13e)

packages/schemas/lib/events/eventTypes.ts
CarlosGamero: Pinning the input parameter closes the scalar-input `transform`/`preprocess` hole. `z.any()` is now rejected explicitly: a `RejectAnyPayload<T>` guard (`IsAny<z.output<T>>`) collapses the payload parameter to `never`. (1fee54a)

packages/schemas/lib/messages/baseMessageSchemas.ts
CarlosGamero: Agreed — the PR is relabeled `minor` and an `UPGRADING.md` entry documents the type-argument arity change (3 → 2; inference call sites are unaffected). (7ff89f5)

packages/schemas/lib/messages/baseMessageSchemas.spec.ts
CarlosGamero: Reworked. Type-level guards now live in `*.types.spec.ts` (the only files vitest type-checks), covering object/union accepted, scalar rejected, and `RejectAnyPayload<any>` → `never` — each verified to fail when the fix is reverted. Runtime `.spec.ts` files assert real parse behaviour. (ec2c846)

CarlosGamero:
Re: the `enrichEventSchemaWithBase` sibling (`baseEventSchemas.ts`) — relaxed it and its `ReturnType` the same way, so all three `enrich*` helpers now share one `EventPayloadSchema` constraint. The shared type helpers (`EventPayloadSchema`, `RejectAnyPayload`) are centralized in `baseEventSchemas.ts`. (4424890)

kibertoad (approved):

PR #586, opened 2026-09-14
Added opt-in retries for event listeners

kjamrog:
Background event listeners were fire-and-forget: a failing handler was logged and reported, never retried. This adds opt-in retries.

## What changed

- `EventHandlerRetryOptions` (`maxRetries`, `baseRetryDelayMs`, `maxRetryDelayMs`, `isRetryable`), configurable per registration. `on`/`onMany`/`onAny` now accept `boolean | { isBackgroundHandler?, retry? }` as their third argument; the boolean form still works but is deprecated and will be removed in the next major release.
- Defaults to `maxRetries: 0` — no behaviour change unless a listener opts in. Bounded at 5 retries and 10s per backoff, and `dispose()` abandons pending backoffs so retries cannot hold shutdown.
- Each attempt gets its own observability transaction. Every failed attempt is logged; only the final failure is rethrown (foreground) or reported to the error reporter (background), now with `attempts` in the log/report context.
- Invalid retry options are rejected when the handler is registered (out-of-range numbers, a non-function or async `isRetryable`, a base delay above the max).
- Non-`Error` throwables are wrapped with the serialized value and `cause` instead of being cast.

## Notes

The background failure log and error report no longer carry the full serialized event. They now carry `eventId`, `eventType`, `eventTimestamp` and `eventMetadata`, but not the payload, which may hold sensitive data. This matches how the rest of the library treats message bodies: `logMessageProcessed` logs ids, timestamps and metadata, and the payload is only logged behind the `logMessages` / `messageLogFormatter` opt-in. Anything consuming `context.event` from these reports needs updating.

Retries happen in-memory within the same dispatch: handlers must be idempotent, nothing is retried after a process restart, and foreground retries stack on top of message-level retries — prefer retrying where `emit()` is called. Documented in the core README.

## Summary by CodeRabbit

* **New Features**
  * Added configurable foreground and background event-handler retries with clearer attempt reporting.
  * Retry failures now preserve non-error thrown values and provide event metadata without exposing payload contents.
  * Added public type exports for retry and registration options.

* **Bug Fixes**
  * Invalid retry settings are now rejected, including excessive retries, unsupported delays, and asynchronous retry predicates.
  * Disposing the emitter abandons pending retry delays while allowing the active attempt to finish.

* **Documentation**
  * Updated retry limits, validation rules, logging details, and shutdown behavior.


packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -263,6 +332,38 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
+        await this.executeHandlerAttempt(event, handler, isBackgroundHandler)
+        return
+      } catch (error) {
+        const attempts = attempt + 1
+        const shouldRetry = attempt < retry.maxRetries && (retry.isRetryable?.(error) ?? true)
+        if (!shouldRetry) {
+          return this.handleFailedEventHandler(event, handler, isBackgroundHandler, error, attempts)
+        }
+
+        this.logger.error({
```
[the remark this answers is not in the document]
kjamrog: Readme adjusted

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -28,6 +30,94 @@ import type {
...
+/**
+ * Retries are performed in-memory, within the same dispatch, so they do not survive a process
+ * restart and the handler must be idempotent. `dispose()` waits for them to finish, so a large
+ * retry budget directly extends graceful shutdown - keep it small.
+ */
+export type EventHandlerRetryOptions = {
+  /**
+   * Number of extra attempts after the initial one. Capped at 10.
+   * @default 0
+   */
```
CarlosGamero, line 62: Should we document the max allowed retries?

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -28,6 +30,94 @@ import type {
...
+   * Base delay for the exponential backoff.
+   * Actual delay = min(baseRetryDelayMs * 2 ^ attempt, maxRetryDelayMs)
+   * @default 100
+   */
+  baseRetryDelayMs?: number
+  /**
+   * Upper bound for the delay between attempts.
+   * @default 1000
+   */
+  maxRetryDelayMs?: number
```
CarlosGamero, line 74: I would probably add a maximun allowed delay here to prevent `dispose` to be stuck

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -28,6 +30,94 @@ import type {
...
+type ResolvedRetryOptions = Required<Omit<EventHandlerRetryOptions, 'isRetryable'>> &
+  Pick<EventHandlerRetryOptions, 'isRetryable'>
+
+const resolveRegistrationOptions = (
+  options: boolean | EventHandlerRegistrationOptions,
+): EventHandlerRegistrationOptions =>
+  typeof options === 'boolean' ? { isBackgroundHandler: options } : options
+
+const resolveRetryOptions = (retry?: EventHandlerRetryOptions): ResolvedRetryOptions => {
+  const resolved = {
```
CarlosGamero: 🟢 Nit: Can be simplified by {...DEFAULT_HANDLER_RETRY_OPTIONS, ...retry}
```suggestion
  const resolved: ResolvedRetryOptions =  {...DEFAULT_HANDLER_RETRY_OPTIONS, ...(retry ?? {})}
```

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -183,14 +278,17 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
   public on<EventTypeName extends EventTypeNames<SupportedEvents[number]>>(
     eventTypeName: EventTypeName,
     handler: SingleEventHandler<SupportedEvents, EventTypeName>,
-    isBackgroundHandler = false,
+    options: boolean | EventHandlerRegistrationOptions = false,
```
CarlosGamero, line 300: This options param might be a bit confusing, maybe we can add some JSDoc to clarify it and rename param to something like `isBackgroundHandlerOrOptions`

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -199,18 +297,21 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
   public onMany<EventTypeName extends EventTypeNames<SupportedEvents[number]>>(
     eventTypeNames: EventTypeName[],
     handler: SingleEventHandler<SupportedEvents, EventTypeName>,
-    isBackgroundHandler = false,
+    options: boolean | EventHandlerRegistrationOptions = false,
```
CarlosGamero, line 321: same here

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -277,25 +436,46 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
+    })
+  }
+
+  private buildErrorContext<SupportedEvent extends SupportedEvents[number]>(
+    event: CommonEventDefinitionConsumerSchemaType<SupportedEvent>,
+    handler: EventHandler<CommonEventDefinitionConsumerSchemaType<SupportedEvent>>,
+    attempts: number,
+  ) {
+    return {
+      event: JSON.stringify(event),
```
CarlosGamero: We can't use `JSON.stringify` on longs, I know it was there before, but could we fix it already? the right method os stringValueSerializer from node-core

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -28,6 +30,94 @@ import type {
...
+
+/**
+ * Retries are performed in-memory, within the same dispatch, so they do not survive a process
+ * restart and the handler must be idempotent. `dispose()` waits for them to finish, so a large
+ * retry budget directly extends graceful shutdown - keep it small.
+ */
+export type EventHandlerRetryOptions = {
+  /**
+   * Number of extra attempts after the initial one. Capped at 10.
+   * @default 0
```
CarlosGamero, line 61: Maybe by default we can have some retries 🤔 what do you think? I am not sure so please share your thoughts, I am also fine as is

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: It's mentioned there: `Capped at 10.`

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Agree, I was actually adding that already :D I think we can be quite strict - setting it to 2 minutes, let me know if it sounds good (on the app side we will use smaller value that fit within shutdown period)

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: I wanted it to be opt-in for now, because we can't assume that all listeners are idempotent - I can imagine cases in which retries might not be expected so I didn't want to change the behavior for existing ones

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: True, I'm actually wondering if we should get rid of boolean variant and create major release with `EventHandlerRegistrationOptions` being the only accepted form. Then if it's major we could also make retries enabled by default as you suggested [here](https://github.com/kibertoad/message-queue-toolkit/pull/586#discussion_r4013427073)

WDYT?

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: ✅

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Good catch, updated

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: Ok ... I am blind, sorry 😮‍💨

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: This is on memory retry by design, so I would say 20-30s is a good number IMO

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: I would avoid the major just for this reason. I would make a note adding that the boolean is deprecated and on next major it will be removed.

packages/core/lib/events/DomainEventEmitter.ts
CarlosGamero: Good point! all good then :D

kibertoad (commented):
Review of the opt-in listener retries. 15 findings, roughly in severity order, posted inline.

The three I would fix before merging:
1. Exhausted foreground retries throw before the background handlers are registered, so they never run and the handler spy never marks the event consumed.
2. The handler-map entry is created before the retry options are validated, so a rejected registration leaves a phantom empty entry that makes the emitter report the event as consumed with zero handlers.
3. The error-report `event` field changed serializer, which breaks consumers that `JSON.parse` it, and is not mentioned in the description.

The rest are configuration and observability gaps (unabortable backoff vs `dispose()`, missing `base <= max` check, retry-everything default, async `isRetryable`, discarded non-Error values) plus a few nits.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -241,12 +365,14 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
       EventHandler<CommonEventDefinitionConsumerSchemaType<SupportedEvents[number]>>
     >,
   ): Promise<void> {
-    for (const handler of eventHandlers.foreground) {
-      await this.executeEventHandler(event, handler, false)
+    for (const registration of eventHandlers.foreground) {
```
kibertoad, line 296: **Foreground retry exhaustion skips every background handler for the event.**

When a foreground handler runs out of retries, `handleFailedEventHandler` rethrows, so `executeHandlers` exits here and the lines below never run: `eventHandlers.background` is never mapped, `inProgressBackgroundHandlerByEventId` is never set, and `_handlerSpy.addProcessedMessage(..., 'consumed')` never fires.

Repro: register one foreground handler with `{ retry: { maxRetries: 3 } }` and one background handler on the same event. After the foreground attempts are exhausted, the background handler receives nothing and `handlerSpy.waitForMessageWithId(id, 'consumed')` hangs until the timeout.

The hole predates this PR, but retries widen it a lot: the foreground handler now fails late rather than immediately, so the window is the whole retry budget. Suggestion: kick off the background handlers before awaiting the foreground ones, and cover the mixed case (foreground fails, background must still run) in a test.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -179,38 +291,50 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
    */
   public on<EventTypeName extends EventTypeNames<SupportedEvents[number]>>(
     eventTypeName: EventTypeName,
     handler: SingleEventHandler<SupportedEvents, EventTypeName>,
-    isBackgroundHandler = false,
+    options: boolean | EventHandlerRegistrationOptions = false,
   ) {
+    const { isBackgroundHandler, retry } = resolveRegistrationOptions(options)
+
     if (!this.eventHandlerMap.has(eventTypeName)) {
```
kibertoad, line 224: **The map entry is inserted before the retry options are validated, leaving a phantom entry that reports the event as consumed with zero handlers.**

`resolveRetryOptions` throws for invalid input, but by then the `{ foreground: [], background: [] }` entry for this event type already exists. `handleEvent` no longer short-circuits on a missing entry, so emitting that type runs `executeHandlers` over two empty arrays and still calls `_handlerSpy.addProcessedMessage(..., 'consumed')`, claiming the event was handled when nothing handled it.

Repro: `emitter.onAny(listener, { retry: { maxRetries: 11 } })` throws, and the entry for the first supported event type survives.

Via `onMany` the same bug also leaves state partially mutated (entry created for event type 1, nothing for 2..n) and re-runs the identical validation once per event type. Suggestion: resolve and validate the options once at the top of `on`/`onMany`, before touching `eventHandlerMap`.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -277,25 +459,46 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
+    })
+  }
+
+  private buildErrorContext<SupportedEvent extends SupportedEvents[number]>(
+    event: CommonEventDefinitionConsumerSchemaType<SupportedEvent>,
+    handler: EventHandler<CommonEventDefinitionConsumerSchemaType<SupportedEvent>>,
+    attempts: number,
+  ) {
+    return {
+      event: stringValueSerializer(event),
```
kibertoad: **Switching this field from `JSON.stringify` to `stringValueSerializer` is a breaking format change for error-report consumers.**

`stringValueSerializer` returns a quoted string (the JSON text wrapped in literal double quotes) and truncates at 8100 chars, so anything downstream doing `JSON.parse(report.context.event)` now fails with `Unexpected non-whitespace character after JSON at position 3`, and large events are silently cut off.

`stringValueSerializer` is the convention elsewhere in the repo (`AbstractQueueService`, kafka), so the direction is fine, but the change is unrelated to retries and is not mentioned in the PR description. Worth a changelog/UPGRADING note, or keeping `JSON.stringify` here so this PR stays additive.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -263,6 +389,62 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
+        }
+
+        this.logger.error({
+          ...resolveGlobalErrorLogObject(error),
+          ...this.buildErrorContext(event, handler, attempts),
+          maxAttempts: retry.maxRetries + 1,
+          msg: `Event handler ${handler.eventHandlerId} failed, retrying`,
+        })
+
+        await setTimeout(Math.min(retry.baseRetryDelayMs * 2 ** attempt, retry.maxRetryDelayMs))
```
kibertoad: **The backoff sleep takes no `AbortSignal`, so `dispose()` can neither interrupt a pending retry nor stop the timer from holding the process open.**

`MAX_ALLOWED_RETRIES` (10) times `MAX_ALLOWED_RETRY_DELAY_MS` (120000) is 20 minutes of sleeping that `dispose()` has to await, which does not match the comment above about keeping the budget small on purpose. A background handler failing permanently with `{ maxRetries: 10, baseRetryDelayMs: 120_000 }` makes `await emitter.dispose()` hang for roughly 20 minutes during a SIGTERM drain, well past a typical orchestrator kill timeout. `timers/promises.setTimeout` is also ref'd by default, so the pending timer keeps the event loop alive even after `dispose()` gives up its drain passes.

Suggestion: an `AbortController` aborted by `dispose()`, passed as `setTimeout(delay, undefined, { signal })`, plus a cap on the aggregate budget rather than only on the per-attempt delay.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -28,6 +32,109 @@ import type {
...
+    !Number.isInteger(resolved.maxRetries) ||
+    resolved.maxRetries < 0 ||
+    resolved.maxRetries > MAX_ALLOWED_RETRIES
+  ) {
+    throw new InternalError({
+      errorCode: 'INVALID_RETRY_OPTIONS',
+      message: `maxRetries must be an integer between 0 and ${MAX_ALLOWED_RETRIES}, received ${resolved.maxRetries}`,
+    })
+  }
+  for (const field of ['baseRetryDelayMs', 'maxRetryDelayMs'] as const) {
```
kibertoad: **No `baseRetryDelayMs <= maxRetryDelayMs` check, so a base above the (defaulted) max is silently flattened instead of rejected.**

`{ maxRetries: 3, baseRetryDelayMs: 5000 }` passes validation: 5000 is under `MAX_ALLOWED_RETRY_DELAY_MS`, and `maxRetryDelayMs` falls back to its 1000ms default. Every delay then computes as `Math.min(5000 * 2 ** attempt, 1000)`, so the caller's intended 5s/10s/20s backoff becomes a flat 1s with no warning.

That is exactly the class of misconfiguration the comment above says this validation exists to catch, so it belongs in the same block.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -28,6 +32,109 @@ import type {
...
+  retry?: EventHandlerRetryOptions
+}
+
+type ResolvedRetryOptions = Required<Omit<EventHandlerRetryOptions, 'isRetryable'>> &
+  Pick<EventHandlerRetryOptions, 'isRetryable'>
+
+const resolveRegistrationOptions = (
+  options: boolean | EventHandlerRegistrationOptions,
+): EventHandlerRegistrationOptions =>
+  typeof options === 'boolean' ? { isBackgroundHandler: options } : options
```
kibertoad: **`null` passes through unchanged (`typeof null === 'object'`), so the destructure in `on` throws a raw `TypeError`.**

A JS caller, or a TS caller holding a `boolean | null`, gets `TypeError: Cannot destructure property 'isBackgroundHandler' of 'null'` with no event type, handler id, or error code attached, while every other bad input on this path surfaces as `InternalError { errorCode: 'INVALID_RETRY_OPTIONS' }`.

Suggestion: `options == null || typeof options === 'boolean' ? { isBackgroundHandler: !!options } : options`.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -263,6 +389,62 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
   private async executeEventHandler<SupportedEvent extends SupportedEvents[number]>(
+    event: CommonEventDefinitionConsumerSchemaType<SupportedEvent>,
+    registration: HandlerRegistration<
+      EventHandler<CommonEventDefinitionConsumerSchemaType<SupportedEvent>>
+    >,
+    isBackgroundHandler: boolean,
+  ) {
+    const { handler, retry } = registration
+
+    for (let attempt = 0; ; attempt++) {
```
kibertoad: **In the foreground, the retry loop blocks `emit()` for the whole budget and can outlast the broker visibility timeout.**

The README says a failing foreground handler propagates so a consumer can retry the whole message, so foreground retries stack on top of the message-level retry the consumer already does. With `{ maxRetries: 10, baseRetryDelayMs: 60_000, maxRetryDelayMs: 120_000 }` (all values pass validation) and a consumer that calls `emit()` inside its message handler, `emit()` is held for up to ~20 minutes. SQS and PubSub redeliver after the visibility timeout (30s by default), so a second consumer starts the same foreground handler while the first is still retrying, and the effective attempt count becomes queue retries times 11.

Suggestion: reject (or ignore) `maxRetries` for foreground registrations, or bound the total budget well below a visibility timeout and document that bound.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -263,6 +389,62 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
+    A throwing predicate must not escape: it would reject the dispatch promise, which for background
+    handlers leaves the in-progress entry behind and surfaces as an unhandled rejection. The handler
+    error it was asked about is the one that matters, so it is left to the regular failure path.
+  */
+  private isRetryable(
+    retry: ResolvedRetryOptions,
+    handler: EventHandler<CommonEventDefinitionConsumerSchemaType<SupportedEvents[number]>>,
+    error: unknown,
+  ): boolean {
+    if (!retry.isRetryable) return true
```
kibertoad, line 376: **The default retries every error, including ones that can never succeed.**

With no `isRetryable` predicate, a handler failing with `InternalError { errorCode: 'EVENT_EMITTER_DISPOSED' }`, `UNKNOWN_EVENT`, or a zod parse failure from a nested `emit()` is retried the full 11 attempts with backoff. Concrete case: once `dispose()` has exhausted its drain passes and set `isDisposed`, a retry loop still in flight calls `emit()`, gets `EVENT_EMITTER_DISPOSED` on every attempt, and sleeps through its entire budget retrying something structurally impossible.

The rest of the repo already separates retryable from permanent failures (`isRetryableGrpcError`, `RetryMessageLaterError`). At minimum, skip retries for emitter-lifecycle and validation errors.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -28,6 +32,109 @@ import type {
...
+      resolved[field] > MAX_ALLOWED_RETRY_DELAY_MS
+    ) {
+      throw new InternalError({
+        errorCode: 'INVALID_RETRY_OPTIONS',
+        message: `${field} must be between 0 and ${MAX_ALLOWED_RETRY_DELAY_MS} ms, received ${resolved[field]}`,
+      })
+    }
+  }
+
+  if (resolved.isRetryable !== undefined && typeof resolved.isRetryable !== 'function') {
```
kibertoad: **The `typeof === 'function'` check accepts an async predicate, which then inverts the predicate's meaning.**

The spec comment notes this runtime check exists for JS callers who ignore the types. A JS caller passing `isRetryable: async (e) => !(e instanceof ValidationError)` clears the check; at dispatch the call returns a pending Promise, which is truthy, so every error counts as retryable, including the ones the predicate was written to exclude. The failure is completely silent.

Suggestion: reject async functions at registration (`predicate.constructor.name === 'AsyncFunction'`), or compare the result with `=== true` and log when it is not a boolean.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -277,25 +459,46 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
+  ) {
+    if (!isBackgroundHandler) throw error
+
+    const context = this.buildErrorContext(event, handler, attempts)
+    this.logger.error({
+      ...resolveGlobalErrorLogObject(error),
+      ...context,
+    })
+    this.errorReporter?.report({
+      error: isError(error)
```
kibertoad, line 439: **The thrown value is discarded, so a non-Error failure becomes unattributable.**

No `cause`, no serialization of the original value. `resolveGlobalErrorLogObject` logs a non-Error as `msg: 'Unknown error'`, so a handler doing `throw 'boom'` produces `{"msg":"Unknown error", ..., "eventHandlerId":"ThrowingListener","attempts":1}` and a report whose error message only says a non-error value was thrown, with a stack pointing back at `handleFailedEventHandler`. The string `boom` appears nowhere, and the same holds for a thrown `{ code: 'X' }` or a rejected non-Error.

`stringValueSerializer` is already imported here: including it in the message and passing `{ cause: error }` keeps the payload.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -263,6 +389,62 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
+      } catch (error) {
+        const attempts = attempt + 1
+        const shouldRetry = attempt < retry.maxRetries && this.isRetryable(retry, handler, error)
+        if (!shouldRetry) {
+          return this.handleFailedEventHandler(event, handler, isBackgroundHandler, error, attempts)
+        }
+
+        this.logger.error({
+          ...resolveGlobalErrorLogObject(error),
+          ...this.buildErrorContext(event, handler, attempts),
```
kibertoad: **Reusing `buildErrorContext` for the per-attempt log repeats the full serialized event on every attempt.**

`buildErrorContext` includes `event: stringValueSerializer(event)` (up to 8100 chars), so with `maxRetries: 10` and a large payload one event emits 11 log lines each carrying up to 8KB of event body plus a full stack. The intermediate line only needs `eventHandlerId`, `attempts`, `maxAttempts`, and the correlation id; the event body is already on the final failure record.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -28,6 +32,109 @@ import type {
...
+} as const
+
+/**
+ * Retries are performed in-memory, within the same dispatch, so they do not survive a process
+ * restart and the handler must be idempotent. `dispose()` waits for them to finish, so a large
+ * retry budget directly extends graceful shutdown - keep it small.
+ */
+export type EventHandlerRetryOptions = {
+  /**
+   * Number of extra attempts after the initial one. Capped at 10.
```
kibertoad: **Docs say the bounds are capped, but out-of-range values throw at registration.**

This line says "Capped at 10", and the README repeats "capped at 10" and "capped at 2 minutes", yet `maxRetries: 20` raises an `InternalError` instead of clamping to 10. For most apps registration happens during DI/bootstrap, so a reader who trusts the docs gets a service that fails to start. `baseRetryDelayMs` below documents no upper bound at all but is validated against 120000, so `baseRetryDelayMs: 300_000` throws unannounced too.

Either clamp as documented, or reword to say values above the limit are rejected and document the base bound.

packages/core/lib/events/DomainEventEmitter.spec.ts
```diff
@@ -431,6 +435,168 @@ describe('DomainEventEmitter', () => {
...
+        {
+          eventHandlerId: 'ThrowingListener',
+          handleEvent: () => {
+            throw 'boom'
+          },
+        },
+        true,
+      )
+
+      const emittedEvent = await eventEmitter.emit(TestEvents.created, createdEventPayload)
```
kibertoad, line 461: **These tests share the module-level `createdEventPayload` fixture, which `emit()` mutates in place.**

`emit()` does `if (!data.id) data.id = ...` on the caller's object, so the first test to emit stamps `id`, `timestamp`, and `metadata` onto the const declared at the top of the file. Running the suite confirms it: every log line across the whole file carries the same event id and correlation id.

The new retry tests only stay correct because `beforeEach` builds a fresh emitter and spy. A test that emits twice, or a switch to a shared emitter, would silently match the wrong record in `waitForMessageWithId`. Suggestion: build a fresh payload per test, for example `{ ...createdEventPayload, metadata: { ...createdEventPayload.metadata } }`.

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -28,6 +32,109 @@ import type {
...
+  baseRetryDelayMs: 100,
+  maxRetryDelayMs: 1000,
+} as const
+
+/**
+ * Retries are performed in-memory, within the same dispatch, so they do not survive a process
+ * restart and the handler must be idempotent. `dispose()` waits for them to finish, so a large
+ * retry budget directly extends graceful shutdown - keep it small.
+ */
+export type EventHandlerRetryOptions = {
```
kibertoad: **This is now the fourth copy of the same retry options shape and backoff formula in the monorepo.**

`packages/gcp-pubsub/lib/pubsub/AbstractPubSubConsumer.ts` (around lines 50-67) already declares `{ maxRetries?, baseRetryDelayMs?, maxRetryDelayMs? }` with the same doc comment about `min(baseRetryDelayMs * 2 ** attempt, maxRetryDelayMs)` and computes it in two places; `packages/core/lib/queues/AbstractQueueService.ts` has another variant. They have already drifted on the exponent: pubsub uses `2 ** (attempt - 1)`, this uses `2 ** attempt`.

Suggestion: export `resolveRetryOptions` and a `calculateBackoffDelay` helper from core and have gcp-pubsub use them, so defaults and clamping live in one place.

packages/core/README.md
```diff
@@ -587,6 +588,37 @@ emitter.on('user.created', async (event) => {
...
+The third argument still accepts a plain boolean (`isBackgroundHandler`) for backwards
+compatibility, and each attempt is reported to the transaction observability manager separately.
+Every failed attempt is logged at `error` level, intermediate ones carrying `attempts` and
+`maxAttempts`; only the final failure is reported to the error reporter, with the number of
+attempts in its context. Invalid retry options are rejected when the handler is registered.
+
+Retries happen in-memory, within the same dispatch:
+
+- handlers must be idempotent, since an attempt may fail after partially completing;
+- nothing is retried after a process restart - work that must not be lost belongs on a queue;
```
kibertoad: Style nit: the spaced hyphen reads as a dash here. A semicolon, a colon, or a second sentence keeps the line cleaner. Same pattern in the `DomainEventEmitter` doc comment about the retry budget extending graceful shutdown.

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Isn't that expected with how `DomainEventEmitter` and handlers are designed? It always worked like that, foreground is a part of `emit()` method effectively - if it fails we don't execute background handlers. 
For that reason I wouldn't add retries on foreground handlers in most cases as the retry should be on higher level where `emit()` is called

packages/core/lib/events/DomainEventEmitter.ts
```diff
@@ -263,6 +332,38 @@ export class DomainEventEmitter<SupportedEvents extends CommonEventDefinition[]>
...
+  ) {
+    const { handler, retry } = registration
+
+    for (let attempt = 0; ; attempt++) {
+      try {
+        await this.executeHandlerAttempt(event, handler, isBackgroundHandler)
+        return
+      } catch (error) {
+        const attempts = attempt + 1
+        const shouldRetry = attempt < retry.maxRetries && (retry.isRetryable?.(error) ?? true)
```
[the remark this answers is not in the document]
kjamrog: Check again

packages/core/lib/events/DomainEventEmitter.ts
kibertoad: let's document that as intentional part of the design then

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: I limited both max delay and retries significantly - it's meant to be few seconds in total

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Already added it in the README

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Done ✅

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Done ✅

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Added ✅

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: I applied stricter limits for both delay and amount of retries - there is also a note in README that retries are mostly designed for background handlers. Foreground handlers are effectively part of `emit()` so the error is propagated outside, therefore application level handling is preferred

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Since this is used for logging and reporting only, we don't consider that a breaking change.

Also, event was removed from error log now to align with how it works in other MQT modules - event payload can contain sensitive information so we are only logging some metadata

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Added JSDoc and renamed the parameter

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Added `ZodError` and internally thrown errors codes as non-retriable. On top of that `isRetryable` allows to add custom logic if needed

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Improved

kibertoad (approved):

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Added both:
- rejecting `AsyncFunction` type 
- logging cases where non-boolean is returned and resolving them to false. This is handle for example `() => Promise.resolve(false)`

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Done ✅

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: I removed the event from the log completely to avoid logging sensitive data (see: https://github.com/kibertoad/message-queue-toolkit/pull/586#discussion_r4025261799)

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: README reworded - we throw if invalid configuration is provided

packages/core/lib/events/DomainEventEmitter.spec.ts
kjamrog: Done ✅

packages/core/README.md
kjamrog: Fixed

packages/core/lib/events/DomainEventEmitter.ts
kjamrog: Adjusted it a bit to be similar to existing convention, but I wouldn't extract it within this PR - it's already big and I would need to touch other packages. I can create a follow-up PR for extracting it

